PR #450, opened 2014-10-02
[WIP] Reimplemented Runtime commands

3y3:
@bajtos , please look at this. This is not ready to review, but I need a discussion here.

What happens here:
- I copy `InjectedScriptSource` and `DebuggerScript` from Blink sources (branch 1625).
This is two main files which implements debugging protocol.
- I add `InjectedScriptHost.js`, this is implementation of some functions stored in `V8InjectedScriptHostCustom.cpp` and used by `injectedScript`
- Next step - I reimplement Runtime agent and some functions in Debugger agent to fix some issues:
  - Now we can inspect:
    - global variables when process not paused
    - get/set properties
    - big arrays (I don't test buffers)
    - function raw scopes
  - RegExp inspection info show `/abc/g` instead of `/abc/`

I use something like this to debug this all:
```js
var bigArray = new Array(99999);
for (var i = 0; i < bigArray.length; i++) bigArray[i] = i;
var rx1 = /abc/gim;
var rx2 = new RegExp('abc', 'gim');
var _var = 0;
var obj = {};
Object.defineProperty(obj, 'prop', {
  get: function() { return _var; },
  set: function(val) { _var = val; }
});
setInterval(function() {
  bigArray; rx1; rx2; obj;
}, 1000);
```

# ...but this commit has two problems
## 1. It is very big
I can try to divide it on parts like `Implemented Debugger.getBacktrace`, `Implemented Runtime.callFunctionOn`, `...`. But I'm not sure about that.
## 2. It deprecates big part of tests
This is a first in big series of pull requests, where I'll reimplement all agents to use WebKit protocol (this is relative to mail discussion, where Brandon Meck promises to implement this protocol in node)
How you can see I deleted some incompatible tests, I would like to see on their place autogenerated tests based on `tools/protocol.json`

Is there something with which you are disagree? Any ideas about changing structure of this commit, which can help you in review?


esco:
:+1:

3y3:
I continue work on this pr #808

Issue #472, opened 2014-11-25
feature request: auto-close if auto-opened


bblack:
It would be convenient, if the debugger was automatically opened in the browser by invoking `node-debug`, to have the debugger automatically `window.close()` when it is detached from the debugger.
##

PR #489, opened 2014-12-09
When inspector url has autoclose=true, then upon detach, frontend closes...

bblack:
... instead of trying to reconnect.

Also, node-debug sets autoclose=true on the url.

addresses #472


bajtos:
Closing as abandoned.

PR #510, opened 2014-12-31
Fix changelog.sh for Windows

3y3:
Previous version doesn't work correctly (it miss first commit message in changelog list).
I can't understand this behavior, but this commit fix it.

### DIFF
--- a/tools/git-changelog.sh
+++ b/tools/git-changelog.sh
@@ -8,9 +8,7 @@ tmp="$TMPDIR/changelog"
 
 if [ "$(expr substr $(uname -s) 1 10)" == "MINGW32_NT" ]; then
   NULL="NUL";
-  # TMP was overriden by Bash (TMP="/tmp") we need to unset it.
-  unset TMP;
-  tmp="$TMP/changelog"
+  tmp="._changelog"
 fi
 
 while [ "$1" != "" ]; do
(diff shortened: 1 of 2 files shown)


3y3:
@bajtos , I'm ready to merge this one and release 0.8.2

Also I'm glad to notice that now we have prebuild binaries for Linux and OSX x64 (I'm ready to add x32 atoreleasing if it's needed)

tools/git-changelog.sh
```diff
@@ -8,9 +8,7 @@ tmp="$TMPDIR/changelog"
 
 if [ "$(expr substr $(uname -s) 1 10)" == "MINGW32_NT" ]; then
   NULL="NUL";
-  # TMP was overriden by Bash (TMP="/tmp") we need to unset it.
-  unset TMP;
-  tmp="$TMP/changelog"
+  tmp="._changelog"
```
bajtos: What is the purpose of this change?

I am not entirely against keeping the temp file in the local directory. However, it must be listed in `.gitignore` in such case, to ensure it is not accidentally committed to git and/or released to npmjs.org. The name `.changelog` would be better in such change.

3y3: First `.gitignore` rule is `._*`, I get the name relative to this rule. Correct me if understand incorrect this mask.

bajtos:
> First .gitignore rule is ._*, I get the name relative to this rule. Correct me if understand incorrect this mask.

Oh, I did not realise we have such rule in `.gitignore`.

Let's not dwell on details and ship this patch then.

LGTM.

PR #511, opened 2014-12-31
Wrong name ScriptManager.normalizeV8Name

3y3:
Oooops! I previous commit we prefer to don't test this part, and I did everything I could properly.
Fixed wrong line.
Added covering test.


bajtos:
LGTM, :shipit:

Issue #506, opened 2014-12-28
Use separate front-end like Chrome DevTools app, instead of including own


auchenberg:
I just released [Chrome DevTools app](https://github.com/auchenberg/chrome-devtools-app), which is a standalone version of Chrome DevTools, independent of Chrome.

![Intro](https://raw.githubusercontent.com/auchenberg/chrome-devtools-app/master/readme/app-intro.png)

I imagine we could make it possible for node-inspector to use [Chrome DevTools app](https://github.com/auchenberg/chrome-devtools-app), as the front-end instead of hosting it's own version?

We could also explore how to broadcast the node-inspector endpoint via mDNS, similar to how [PonyDebugger](https://github.com/square/PonyDebugger) is making endpoints discoverable by the front-end. 

How about that?
##

3y3:
Are you propose to completely deprecate `frontend` folder in Node Inspector project?
If it's so, then it's impossible, because Inspector's frontend has some specials.

bajtos:
Also keep in mind that the frontend is tightly coupled with the protocol version implemented by the backend. For example, the current Node Inspector version implements protocol from 09/2013. Recent versions of dev-tools frontend may require new commands and/or different response format, depending on the changes made to the protocol in the past.

Having said that, I am open to experimentation as long as the integration with Node Inspector is lightweight and don't add extra maintenance for us.

auchenberg:
@3y3 Yes I'm proposing to have a "generic" DevTool app, that is able to work with multiple targets, ex. chrome, firefox, node-inspector, PonyDebugger, etc.  The point here is that each "back-end provider" shouldn't include their own front-end, but focus on providing a backend that is compliant to the Chrome Remote Debugging protocol.

What are the special needs of the node-inspector front-end? It's my understanding that you are using a pull of DevTools from blink, where you are overriding a few configurations, like toggling certain tabs. We should be able to find a way to avoid these overrides.

@bajtos That's a good point. With the latest pull of DevTools from blink master, I'm seeing a few "unsupported errors" exceptions being thrown, but that should be solvable too.

3y3:
I would have postponed the issue for at least six months.
Before this discussion I prefer to update frontend, inject DevTools protocol in app, make Node Inspector more extensible, add gui configuration.

If after this steps will be possible to use external frontend, I will be glad to exclude front-end folder and will vote +1 on this issue.

bajtos:
[...]

PR #515, opened 2015-01-05
WIP: Register node-inspector as a RemoteDebug-endpoint over mDNS/Bonjour

auchenberg:
This PR makes it possible for external tools like [chrome-devtools-app](https://github.com/auchenberg/chrome-devtools-app/), to discover an instance of node-inspector as a "RemoteDebug-endpoint" over mDNS/Bonjour, as a follow up on https://github.com/node-inspector/node-inspector/issues/506. 

The idea here is to enable discovery of [RemoteDebug](http://remotedebug.org)-compliant endpoints over MDNS. The metadata is based upon the existing Chrome Remote Debug HTTP API. I'm planning to write a little spec for RemoteDebug Endpoint MDNS functionality too. 

bonjour browser
### Changes:
- [x] New utility methods has been added to generate URL's.
- [x] Added remoteDebugService to register the service over MDNS.
- [x] Change default web-host to local detected IP instead of 127.0.0.1
### Questions:

1) I had to change the default web-host to the local IP instead of 127.0.0.1, as the IP is needed in order to enable other clients to connect. Does this open up a security risk? Is there any reason why 127.0.0.1/0.0.0.0 currently is used?


3y3:
@bajtos , please join to discussion.

@auchenberg , thanks for contribution. This is interesting feature.

>  Is there any reason why 127.0.0.1/0.0.0.0 currently is used?

From time to time I know why we use `0.0.0.0` in `NODE_INSPECTOR_MODE` and `127.0.0.1` in `NODE_DEBUG_MODE`, but unfortunately now I can't remember this.

> I had to change the default web-host to the local IP instead of 127.0.0.1, as the IP is needed in order to enable other clients to connect.

Can you explain me more verbose this part. At first glance new IP detecting opens weak security risk. I prefer to have minimal network area in default configuration, but it's discussable.

lib/config.js
```diff
@@ -435,7 +436,6 @@ function collectDefaultsFromDefinitions(NODE_DEBUG_MODE) {
   });
 
   if (NODE_DEBUG_MODE) {
-    options['web-host'] = '127.0.0.1';
```
bajtos: Please keep `127.0.0.1` for `NODE_DEBUG_MODE`. This mode is triggered when debugging via `node-debug app.js`, in which case the front-end is immediately opened in the browser.

package.json
```diff
@@ -35,7 +35,9 @@
     "which": "^1.0.5",
     "v8-debug": "~0.3.0",
     "v8-profiler": "~5.2.0",
-    "semver": "^3.0.1"
+    "semver": "^3.0.1",
+    "mdns": "~2.2.2",
```
bajtos: Please use `^2.2.2`.

bajtos:
> I had to change the default web-host to the local IP instead of 127.0.0.1, as the IP is needed in order to enable other clients to connect. Does this open up a security risk? Is there any reason why 127.0.0.1/0.0.0.0 currently is used?

127.0.0.1 is used in node-debug mode, as it is the most sensible value. The user is loading the frontend in a local browser, thus a local loopback address is good enough for making the connection, and there is no need to open the debugger server to the whole network.

When running `node-inspector` in a standalone mode, 0.0.0.0 is used mostly for historical reasons. I don't mind changing it to 127.0.0.1, as that is more secure. However, I am opposed to changing it to network IP address as proposed here.

IIUC mDNS/Bonjour correctly, "it provides a general method to discover services on a local area network" ([wikipedia](http://en.wikipedia.org/wiki/Bonjour_%28software%29)), which means that we need to register with an IP address visible from the local network.

I am proposing the following changes to your patch:
 - Keep the configuration of listening address intact, i.e. 127.0.0.1/0.0.0.0.
 - Modify the code registering with mDNS and make it smarter. Detect the situation when the server is listening on 0.0.0.0 and replace 0.0.0.0 with the local network address.

auchenberg:
Good points @bajtos. 

My intention is to enable the MDNS broadcasting out of the box, in order to minimize the workflow, but I understand the added security risk of binding to a addressable IP. 

Alternatively we could add a command-line flag that would enable MDNS broadcasting, the user could flip when running ``node-debug`` and ``node-inspector``? That way it's explicit. 

Flag enabled: ``web-host`` is set to the local addressable IP, and MDNS broadcasting is enabled

bajtos:
I am ok with enabling MDNS broadcasting always, or alternatively when Node Inspector is listening on a public address.

Here is what I am proposing:
 - `node-debug app.js` keeps listening on 127.0.0.1, this command is not intended for debugging via chrome-devtools-app.
 - `node-inspector app.js` keeps listening on 0.0.0.0, a public address is registered with MDNS.

auchenberg:
Changes implemented:

- [x] node-debug app.js keeps listening on ``127.0.0.1``, this command is not intended for debugging via chrome-devtools-app.

- [x] When ``web-host`` is set to ``0.0.0.0``, the local addressable IP is detected, and used for websocket, inspector and frontend URL generation, plus MDNS broadcast is enabled. 

I still have 3 tests failing, but these are also failing for me in mester. Any cure?

3y3:
Are you using node v0.10.35?

auchenberg:
I had 3 tests failing on v0.10.33, and now with 0.10.35, 29 tests are failing.

lib/debug-server.js
```diff
@@ -78,6 +94,11 @@ DebugServer.prototype._createSession = function(debugPort) {
   return Session.create(debugPort, this._config);
 };
 
+DebugServer.prototype._shouldAnnounceEndpoint = function() {
+  var addressInfo = this._httpServer.address();
+  return addressInfo.address === '0.0.0.0';
```
bajtos: I think we should support any non-localhost IPs too.

http://tools.ietf.org/html/rfc5735
```
127.0.0.0/8 - This block is assigned for use as the Internet host
loopback address.  A datagram sent by a higher-level protocol to an
address anywhere within this block loops back inside the host.  This
is ordinarily implemented using only 127.0.0.1/32 for loopback.  As
described in [RFC1122], Section 3.2.1.3, addresses within the entire
127.0.0.0/8 block do not legitimately appear on any network anywhere.
```

A proposed condition:

```js
return !/^127\.0\.0\./.test(addressInfo.address)
```

Please include a descriptive comment linking to the RFC.

lib/debug-server.js
```diff
@@ -85,11 +106,40 @@ DebugServer.prototype.close = function() {
...
 };
 
-DebugServer.prototype.address = function() {
-  var address = this._httpServer.address();
+DebugServer.prototype.getHttpServerAddressInfo = function() {
+  var addressInfo = this._httpServer.address();
+
+  if(this._shouldAnnounceEndpoint()) {
+    // Resolve local IP, in order for computers in the network to resolve
+    addressInfo.address = localIp();
```
bajtos: Local IP should be used only when the address is `0.0.0.0`. When the user explicitly specifies a public IP, then this IP should be honoured in the broadcasted event.

auchenberg:
I'm holding on this PR, until I find a suitable replacement for ``mdns``, as we can't add such a dependency to the project.

bajtos:
Closing as abandoned.

Issue #346, opened 2014-04-09
MAC: node-debug should not open with default browser if it is Firefox


rick-kilgore:
If the user has set Firefox as their "default" browser, then your use of opener in node-debug to open the browser is a bad choice, since node-inspector doesn't work in Firefox.  It should be smart enough to choose Safari instead, or better yet, check try running Chrome, and if that fails, then run Safari, and if that fails then print an error message saying it couldn't find and run Chrome or Safari.

bajtos:
See #528 for a better implementation (in progress).

Issue #345, opened 2014-04-09
Default browser is always Safari on MAC


dotnetwise:
Regardless of what is your default browser `node-debug` always open the inspector page on safari :-1: 
That sucks

PR #528, opened 2015-01-16
Make sure to open URL in a webkit browser

jakub-g:
[commit message]

`opener` always opens the URL in the default browser of the user.
It might be annoying for a user whose default browser is Firefox
to have Node Inspector start in Firefox.

This PR makes sure that the Node Inspector URL is opened
either in Chrome, Chromium or Opera (if one of them is the default
browsers, it gets priority). If we can't do that for some reason
(e.g. user has none of them installed), we just display an error
message in the console but do not open the URL in any browser.

---------------------

Hello guys,

This PR is quite small but takes advantage of a couple of modules for detecting available browsers [4, 5], default browser [2, 3], and launching them [1, 4]. I wrote [1] and [2] myself because I couldn't find any module doing the thing that was cross-platform.

Known issues with [1] for now are the following:
- unable to close the opened browser (but you didn't do it before either)
- it doesn't detect when Chromium is installed on Windows (because it depends on [4] which depends on old version of [5]; the bug is fixed in [5]) - but detects Opera and Chrome.

Let me know what do you think about this PR!

Also, I'm looking forward to the reports from as many Linuxes as possible whether

```
npm install -g x-default-browser
x-default-browser
```

correctly reports the default browser (it might not report some exotic browsers, PRs welcome).

Probably it's not a big deal for most of you who have Chrome as a default browser, but it's pretty annoying for me having Firefox as a default :)

[1] https://github.com/jakub-g/biased-opener
[2] https://github.com/jakub-g/x-default-browser
[3] https://github.com/sindresorhus/default-browser-id
[4] https://github.com/benderjs/browser-launcher2
[5] https://github.com/vweevers/win-detect-browsers

### DIFF
--- a/bin/node-debug.js
+++ b/bin/node-debug.js
@@ -5,7 +5,7 @@ var fork = require('child_process').fork;
 var fs = require('fs');
 var path = require('path');
 var util = require('util');
-var open = require('opener');
+var open = require('biased-opener');
 var whichSync = require('which').sync;
 var inspector = require('..');
 
@@ -194,7 +194,17 @@ function openBrowserAndPrintInfo() {
   );
 
   if (!config.options.cli) {
-    open(url);
+    // try to launch the URL in one of those browsers in the defined order
+    // (but if one of them is default browser, then it takes priority)
+    open(url, {
+        preferredBrowsers : ['chrome', 'chromium', 'opera']
+      }, function(err, okMsg) {
+        if (err) {
+           // unable to launch one of preferred browsers for some reason
+           console.log(err.message);
+           console.log('Please open the URL manually in Chrome/Chromium/Opera or similar browser');
+        }
+    });
   }
 
   console.log('Node Inspector is now available from %s', url);
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "strong-data-uri": "~0.1.0",
     "debug": "^1.0",
     "ws": "~0.4.31",
-    "opener": "^1.3.0",
+    "biased-opener": "~0.2.2",
     "yargs": "^1.2.1",
     "which": "^1.0.5",
     "v8-debug": "~0.3.0",


bin/node-debug.js
```diff
@@ -194,7 +194,19 @@ function openBrowserAndPrintInfo() {
...
+    // try to launch the URL in one of those browsers in the defined order
+    // (but if one of them is default browser, then it takes priority)
+    open(url, {
+        preferredBrowsers : ['chrome', 'chromium', 'opera']
+      }, function(err, okMsg) {
+        if (err) {
+           // unable to launch one of preferred browsers for some reason
+           console.log(err.message);
+           console.log('Please open the URL manually in Chrome/Chromium/Opera or similar browser');
+        } else if (okMsg) {
```
3y3: Is there a situation when `err == null` and `okMsg == null`?

3y3:
Sorry for delay in review.
This is a good stuff, I want to see it in node-inspector. I'm worried about stability of `biased-opener`, but how I can see `opener` itself also hasn't test coverage, so this replacement is equivalent in this context.

/cc @bajtos

bin/node-debug.js
```diff
@@ -194,7 +194,19 @@ function openBrowserAndPrintInfo() {
   );
 
   if (!config.options.cli) {
-    open(url);
+    // try to launch the URL in one of those browsers in the defined order
+    // (but if one of them is default browser, then it takes priority)
+    open(url, {
+        preferredBrowsers : ['chrome', 'chromium', 'opera']
+      }, function(err, okMsg) {
```
3y3: What types of errors can happen here?

bajtos:
This is awesome :star2: :star2: :heart: 

@3y3 I don't see any major problems in this patch, I'll leave the review up to you.

bajtos:
Once the patch is landed, we can probably close the following issue: https://github.com/node-inspector/node-inspector/issues/346

BTW what happens when none of the preferred browsers was found?

bajtos:
> BTW what happens when none of the preferred browsers was found?

NVM, you have already answered that in the PR description:

> If we can't do that for some reason
(e.g. user has none of them installed), we just display an error
message in the console but do not open the URL in any browser.

jakub-g:
Hi, thanks for initial review; I'll come back to this later today or over the weekend; I'll probably also need to do some changes to accomodate for issue reported in #345 when running as sudo

bin/node-debug.js
jakub-g: In my code there are [4 cases of calling `cb` directly](https://github.com/jakub-g/biased-opener/blob/master/index.js#L83-149) which is either `cb(new Error)` or `cb(null, String)`, and one more is passing `cb` to `opener` which invokes [`child_process.execFile(cb: function (error:Error, stdout:Buffer, stderr:Buffer))`](http://nodejs.org/api/child_process.html#child_process_child_process_execfile_file_args_options_callback)

In my tests when I `console.dir(stdout)` in opener's callback it's always an empty string.

However it seems that success callback on Ubuntu (when going through `opener`, i.e. when opening default browser) is called **after the browser is closed** (on Windows it's after it's opened). When going through `browser-launcher`, the callback success is called always **immediately after the browser is opened**.

I may always use `browser-launcher` instead of using `opener` when opening the default browser, I was not sure which way to go actually. Any preferences about it?

bin/node-debug.js
jakub-g: 1. misconfiguration (will not happen here)
2. chrome/chromium/opera not found in the system
3. error from browser-launcher, unable to start the executable for some reason (not sure in which conditions this could happen; probably with improper rwx permissions on unixes and maybe some other circumstances)

(look for `new Error` in https://raw.githubusercontent.com/jakub-g/biased-opener/master/index.js)

jakub-g:
Regarding #345: As far as I understand, the issue reported there would not happen here, since when run with `sudo`, I would first query as the root for the default browser, if this returns safari, I would then query for all browsers to see if there's opera/chrome/chromium installed.

jakub-g:
Updated the PR, regarding the `okMsg` I thought it's always better to have some more logs, which can be useful in case of bug reports, but it's fine for me to skip it.

3y3:
LGTM. Merged.
Thanks for this pr!

Issue #498, opened 2014-12-17
node-debug very delayed source file hydration


samccone:
I am running node-inspector 0.8.1

The command I am running is
`node-debug --hidden='[\"node_modules\"]' -–no-preload --debug-brk=0 `

After around 2 min the sources tab fills in with my files.

Even though I specify node_modules as hidden the node_modules are still showing in the sources.

![screen shot 2014-12-17 at 4 54 52 pm](https://cloud.githubusercontent.com/assets/883126/5479683/8741144a-860d-11e4-9450-bb02844fa464.png)

I suspect this is the real cause of the slow boot up time. Any insight would be greatly appreciated, and thanks so much for this great tool!

bajtos:
It seems like `node-debug` is not correctly forwarding `--hidden` option to Node Inspector.

samccone:
awesome! Let me know if I can help out at all in debugging

samccone:
Alright after using the work done in #530 it seems like this problem is still around

![screen shot 2015-01-22 at 10 33 08 am](https://cloud.githubusercontent.com/assets/883126/5858566/2c773ea0-a222-11e4-88c5-beb237c0b6b5.png)

As you can see it takes around 2 min for the UI with my files to load!

After looking around a bit... this line seems highly suspicious 

https://github.com/node-inspector/node-inspector/blame/813354c66e1fee70012aae94863ead0d3eb68abe/front-end/utilities.js#L1297

thoughts?

bajtos:
My take is that the line 1297 is where the front-end is loading the content of all source files in your project. I guess Chrome developers assume that the files (scripts) are already loaded in the browser, thus its cheap and fast to get them from the backend. Unfortunately that's not the case of Node Inspector, which has to load all of them from the filesystem.

I see two ways how to work around that:
 - reduce the number of files served to front-end (e.g. via `--hidden="node_modules"`)
 - contribute a change to Chromium DevTools front-end and make it load files lazily, only after the user asked front-end to open the file

jldec:
I had a similar problem (2-5 minute delays on debugger startup) and did a little digging.

The issue in my case came from giant glob results (cost ~40s) and subseqent filtering (cost ~2 min!) on the backend.

[...]

PR #529, opened 2015-01-17
Update README.md

zwhitchcox:
Add exclusive include to make sure people know how they can include only certain files. This is extremely useful if you're trying to debug a certain file.

I also changed one of the examples on line 340, because the options have to be specific before the main script, or they will not be included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,19 @@ Just make sure port 8080 is accessible.
 
 Create a JSON-encoded array.  You must escape quote characters when using a command-line option.
 
+#### My options are not being considered
+
+Make sure you put the options before the main script argument.
+
+Right:
+```
+node-debug --hidden=test.js app
+```
+Wrong:
+```
+node-debug app --hidden=test.js
+```
+
 ```sh
 $ node-inspector --hidden='["node_modules/framework"]'
 ```
@@ -337,7 +350,7 @@ $ node-debug --web-host 127.0.0.2 app
 ``` 
 Pass `--option=value` to debugging process:
 ```
-$ node-debug app --option value
+$ node-debug --option value app
 ```
 Start node-inspector to listen on HTTPS:
 ```
@@ -346,6 +359,14 @@ $ node-debug --ssl-key ./ssl/key.pem --ssl-cert ./ssl/cert.pem app
 Ignore breakpoints in files stored in `node_modules` folder or ending in `.test.js`:
 ```
 $ node-debug --hidden node_modules/ --hidden \.test\.js$ app
+```
+Only show files with the word 'test.js' in the path:
+```
+$ node-debug --hidden=^\(?\!.*test\.js.*\) app
+```
+Only show files with the word 'test.js' or 'hello.js' in the path:
+```
+$ node-debug --hidden=^\(?\!\(.*test\.js.*\)\|\(.*hello\.js.*\)\) app
 ```
 Add `--harmony` flag to the node process running the debugged script:
 ```


3y3:
Hello, @zwhitchcox , are you sure that you don't have any problems with `node-debug ... --hidden=... app` ?
This question is relative to #498

zwhitchcox:
Actually, it has stopped working since the last commit.

3y3:
I opened a PR #530 
Can you please confirm, that it fixes a bug.

zwhitchcox:
runnin' like a dream!

README.md
```diff
@@ -337,7 +350,7 @@ $ node-debug --web-host 127.0.0.2 app
 ``` 
 Pass `--option=value` to debugging process:
 ```
-$ node-debug app --option value
+$ node-debug --option value app
```
3y3: Approved change

README.md
```diff
@@ -347,6 +360,14 @@ Ignore breakpoints in files stored in `node_modules` folder or ending in `.test.
...
 $ node-debug --hidden node_modules/ --hidden \.test\.js$ app
 ```
+Only show files with the word 'test.js' in the path:
+```
+$ node-debug --hidden=^\(?\!.*test\.js.*\) app
+```
+Only show files with the word 'test.js' or 'hello.js' in the path:
+```
+$ node-debug --hidden=^\(?\!\(.*test\.js.*\)\|\(.*hello\.js.*\)\) app
+```
```
3y3: I prefer to remove this change, because it is a verbose explanation of 
>Ignore breakpoints in files stored in `node_modules` folder or ending in `.test.js`:
 ```
 $ node-debug --hidden node_modules/ --hidden \.test\.js$ app
 ```

Reasonable to extend extend existed part to something like:

>Ignore breakpoints in files stored in `node_modules` folder or ending in `.test.js`
Hide this files in sources tree:
 ```
 $ node-debug --hidden node_modules/ --hidden \.test\.js$ app
 ```

README.md
```diff
@@ -129,6 +129,19 @@ Just make sure port 8080 is accessible.
...
+
+Right:
+```
+node-debug --hidden=test.js app
+```
+Wrong:
+```
+node-debug app --hidden=test.js
+```
+
```
3y3: This change is not so important to add it here.
I don't like to extend `Troubleshooting` any time. This part is already too big and prevents users see useful `config` info.

We have
>
### How do I specify files to hide?
Create a JSON-encoded array. You must escape quote characters when using a command-line option.
```
$ node-inspector --hidden='["node_modules/framework"]'
```

I prefer to make small update in this part - add `node-debug` example,

README.md
bajtos: This looks suspicious to me. IIRC the arguments *before* the script file are processed by Node Inspector and/or node-debug. Only the arguments *after* the script file are passed to the debugged node process. @3y3: am I wrong?

README.md
bajtos: I like the idea behind these  new examples as they shows how to implement a white-list using the black-list-based option `--hidden`.

However, instead of escaping each character, the example should enclose the regexp in single quotes, e.g.

```
--hidden='^(?!.*test\.js.*)'
'--hidden=^(?!.*test\.js.*)'
```

README.md
bajtos: AFAIU, this example is illustrating how to pass arguments to the process being debugged. The proposed version shows how to pass arguments to Node Inspector instead, which is already covered by the previous examples with `-p` and `--web-host`.

Having said that, this is IMO a good place where to include the note "Make sure you put the Node Inspector options before the main script argument."

3y3: Sorry, I'm wrong here.

PR #530, opened 2015-01-18
Fix --hidden option for NODE_DEBUG mode

3y3:


bajtos:
The change LGTM. Could you please check why the tests are failing on Travis? I had restarted v0.10 job to see if it was a timing issue only.

3y3:
Travis uses 0.10.35 with corrupted debugger

bajtos:
Oh, I see. Is it worth chaning `.travis.yml` to force Node v0.10.33 and thus fix the build?

3y3:
> Oh, I see. Is it worth chaning .travis.yml to force Node v0.10.33 and thus fix the build?

To fix all tests we need to use 0.10.33 and 0.11.13 (current 0.11.14 config downloads prebuild binaries for 0.11.13 and fails. I'll fix this issue in next `v8-debug` version. So we can't publish prebuild binaries for 0.11.14 - there is node-pre-gyp issue with Win x64 build)

bajtos:
I am more concerned about v0.10. I am ok with ignoring failing 0.11 tests for now, as long as we have at least single platform that is passing...

bajtos:
BTW we should probably deprioritize support for v0.11 and focus more on io.js (a.k.a node 1.0). Unfortunately Travis does not support it OOTB, see https://github.com/iojs/io.js/issues/364 and https://github.com/creationix/nvm/issues/590

samccone:
:+1: this fixes #498

bajtos:
@3y3 let's not worry about the CI in this PR and ship the fix to the users. Do you have time to land & release this PR, or shall I do it myself?

bajtos:
The CI build will be fixed by #532.

PR #533, opened 2015-01-23
Fix: Breakpoints doesn't work for Windows

3y3:
Wrong test + wrong realisation = breakpoints doesn't work on windows in main file.
Issue fixed. One test fixed.


bajtos:
Looks good, thanks for fixing it :bow: 

:shipit:

Issue #370, opened 2014-05-12
Cannot set breakpoints in module loaded with npm link


overlookmotel:
I'm developing a module which sits in a folder outside of the application that I'm debugging. i.e.:
Application is at `/work/the-app/`
Module is at `/work/modules/the-module/`

I include the module into the app using `npm link`, which creates a symlink from `/work/modules/the-module/` to `/work/the-app/node_modules/the-module/`

When debugging with node-inspector, breakpoints set in the linked module are skipped over. Is it possible to debug the module like this?

Mac OS X Mountain Lion 10.8.5, node v0.10.26, node-inspector v0.7.3

bajtos:
FWIW, the glob module can follow symlinks since v4.5.0 and v5.0.0, see https://github.com/isaacs/node-glob/pull/148.

sonewman:
Nice :+1:

gxela:
This is still an issue, I am not able to debug a node module that was npm link'd into the project, the project is run from docker containers. Am able to debug node.js app running in docker container with vscode and WebStorm, but not able to set a breakpoint in code that is npm link'd into the project. Let me know if you need an example dockerfile, I can write one up for you. This issue should still be open because it's still an issue.

iyobo:
This was working for me until just last week all of a sudden. had been using Webstorm 2017 all year and it was working. Never upgraded. It just....stopped...using breakpoints in npm-linked moduled.

Breakpoints stop in node_modules that are NOT npm-linked however....so there's definitely something weird about how it reads links.

Perhaps it might even be an OS level "update" change that screwed things up (osx)? I'm still investigating.

huiwang:
I resolved this by running node with --preserve-symlinks

PR #550, opened 2015-02-05
bump glob version, add -sdir option for #498

jldec:
see issue #498 
when there are symlinked directories debugger may take minutes to load

the latest version of glob does not search symlinked directories but allowing this is sometimes desireable
( https://github.com/isaacs/node-glob/issues/134 )

using the `*/**/**` pattern, makes at least one extra level of symlinked directories accessible for debugging but is quite a lot slower, so i added a config option to turn this on only when required.

tests are passing, and delays on deep symlinked projects are now acceptable (~5s instead of ~5min).

refs:
- https://github.com/adgad/express-handlebars/commit/803129ed9b325640cc527aefdb5096dda7661355
- https://github.com/ericf/express-handlebars/pull/98

### DIFF
--- a/lib/ScriptFileStorage.js
+++ b/lib/ScriptFileStorage.js
@@ -2,6 +2,8 @@ var fs = require('fs');
 var path = require('path');
 var async = require('async');
 var glob = require('glob');
+var debug = require('debug')('node-inspector:ScriptFileStorage');
+
 
 var MODULE_HEADER = '(function (exports, require, module, __filename, __dirname) { ';
 var MODULE_TRAILER = '\n});';
@@ -13,6 +15,7 @@ var MODULE_WRAP_REGEX = new RegExp(
 
 var CONVENTIONAL_DIRS_PATTERN = '{*.js,lib/**/*.js,node_modules/**/*.js,test/**/*.js}';
 var ALL_JS_PATTERN = '**/*.js';
+var ALL_JS_PATTERN_SDIR = '**/**/*.js'; // glob extra level of symlinked dirs (slower)
 
 function escapeRegex(str) {
   return str.replace(/([/\\.?*()^${}|[\]])/g, '\\$1');
@@ -27,6 +30,7 @@ function ScriptFileStorage(config, scriptManager) {
   config = config || {};
   this._scriptManager = scriptManager;
   this._noPreload = config.preload === false;
+  this.sdir = config.sdir;
 }
 
 var $class = ScriptFileStorage.prototype;
@@ -148,6 +152,7 @@ $class.listScripts = function(rootFolder, pattern, callback) {
   //   { root: rootFolder },
   //    callback
   // );
+  debug('glob: %s %s', pattern, rootFolder);
 
   glob(
     pattern,
@@ -185,7 +190,9 @@ $class._findScriptsOfRunningApp = function(mainScriptFile, callback) {
     [
       this.findApplicationRoot.bind(this, mainScriptFile),
       function(dir, isRoot, cb) {
-        var pattern = isRoot ? ALL_JS_PATTERN : CONVENTIONAL_DIRS_PATTERN;
+        var pattern = isRoot ?
+          (this.sdir ? ALL_JS_PATTERN_SDIR : ALL_JS_PATTERN) :
+          CONVENTIONAL_DIRS_PATTERN;
         this.listScripts(dir, pattern, cb);
       }.bind(this)
     ],
@@ -200,7 +207,9 @@ $class._findScriptsOfStartDirectoryApp = function(startDirectory, callback) {
       if (!result) {
         callback(null, []);
       } else {
-        this.listScripts(startDirectory, ALL_JS_PATTERN, callback);
+        this.listScripts(startDirectory,
+          (this.sdir ? ALL_JS_PATTERN_SDIR : ALL_JS_PATTERN),
+          callback);
       }
     }.bind(this)
   );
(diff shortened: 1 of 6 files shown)


lib/ScriptFileStorage.js
```diff
@@ -224,12 +233,16 @@ $class.findAllApplicationScripts = function(startDirectory, mainScriptFile, call
     function(err, results) {
       if (err) return callback(err);
 
+      debug('findAllApplicationScripts filtering start');
+
       var files = results[0].concat(results[1]);
       // filter out duplicates and files to hide
       files = files.filter(function(elem, ix, arr) {
         return arr.indexOf(elem) >= ix && !this._scriptManager.isScriptHidden(elem);
```
bajtos: I suspect that `arr.indexOf` can be the root cause of low performance, as this is an `O(n^2)` algorithm.

bajtos:
Hi @jldec, thank you for the pull request. 

In https://github.com/node-inspector/node-inspector/issues/498#issuecomment-72928660, you write:

> The issue in my case came from giant glob results (cost ~40s) and subseqent filtering (cost ~2 min!) on the backend.

And here you say

> delays on deep symlinked projects are now acceptable (~5s instead of ~5min).

I'd like to understand better how did we arrive at 5 seconds. Is it because you run node-debug with symlinks disabled? Or is it because new glob version with `**/**/*` is faster than the old glob version following all symlinks?

What is the breakdown of glob + filtering times in your project with this patch applied and when symlinks are and are not followed?

In general, I try to make Node Inspector work out of the box for as many projects as possible and keep the number of configuration options low. If we can get a similar improvement by fixing the filtering algorithm instead of disabling symlink follows, then I would much more prefer that solution.

---

Another important fact related to symlinks and `findAllApplicationScripts`: The only reason why we try to find all scripts at the beginning is to allow NI users to set breakpoints in files that are not loaded in V8. However, when you have symlinks in your project, these breakpoints don't work anyway, because UI shows a symlink while V8 uses the resolved real path - see #370.

If you don't need to set breakpoints in files that are not loaded in V8 yet, then you can run Node Inspector with `--no-preload`: it disables `findAllApplicationScripts` and thus the startup should be pretty quick.

jldec:
Hi @bajtos 
simply using the latest 'glob' module stops symlink exploration and there is no way to re-enable it. This is probably fine for most users of NI. With just that change i am able to start NI in under a second on my project, but i can't see files in directories below modules which are local via `npm link`. 

the `/**/**/*` pattern is a slightly hackish solution which tells glob to search one extra level of directories (even inside symlinked modules) but i does slow things down. Most NI users won't need it, so i thought it would be best to make it optional with the `-s` option.

this PR also adds a little debug instrumentation under the name `node-inspector:ScriptFileStorage`. Here is a example of the output without `-s`.

```
$ DEBUG=node-inspector:ScriptFileStorage node-inspector
Node Inspector v0.8.3
Visit http://127.0.0.1:8080/debug?port=5858 to start debugging.
  node-inspector:ScriptFileStorage glob: **/*.js /Users/hello/composer/composer +0ms
  node-inspector:ScriptFileStorage findAllApplicationScripts filtering start +614ms
  node-inspector:ScriptFileStorage findAllApplicationScripts filtering done. +135ms
```

The glob took ~614ms and the filter ~135ms

With `-s` the output looks like this:

```
$ DEBUG=node-inspector:ScriptFileStorage node-inspector -s
Node Inspector v0.8.3
Visit http://127.0.0.1:8080/debug?port=5858 to start debugging.
  node-inspector:ScriptFileStorage glob: **/**/*.js /Users/hello/composer/composer +0ms
  node-inspector:ScriptFileStorage findAllApplicationScripts filtering start +4s
  node-inspector:ScriptFileStorage findAllApplicationScripts filtering done. +279ms
```

In this case the glob took ~4s and the filter 279ms

jldec:
for comparison, here are the timings with the older 'glob' module (v3.2.11) which recurses symlink directories.

```
$ DEBUG=node-inspector:ScriptFileStorage node-inspector
Node Inspector v0.8.3
Visit http://127.0.0.1:8080/debug?port=5858 to start debugging.
  node-inspector:ScriptFileStorage glob: **/**/*.js /Users/hello/composer/composer +0ms
  node-inspector:ScriptFileStorage findAllApplicationScripts filtering start +1m
  node-inspector:ScriptFileStorage findAllApplicationScripts filtering done. +4m
```

in this case the glob takes 1 min and the filter 4 min.

jldec:
Regarding how this relates to #370

I concede that allowing debugger access to directories in symlinked modules is not that useful if breakpoints won't work for js files in those directories. 

So until #370 is fixed, i would suggest simply updating NI to use the latest glob and leave it at that. This would at least protect against the circular symlink trap.

bajtos:
@jldec I am still not convinced that the root cause is in symlinks. In my opinion, the problem is in the high number of files in node_modules dependencies.

What happens if you `npm pack && npm install` your dependencies instead of `npm link`? Will this pull request fix the startup time for you? My guess is that it won't.

I would like to improve the current `--preload`/`--no-preload` config option to take a glob pattern specifying which files to preload. This pattern can be `!node_modules/**/**/*` by default.

> So until #370 is fixed, i would suggest simply updating NI to use the latest glob and leave it at that. This would at least protect against the circular symlink trap.

Circular symlink trap - is that the reason why the current version of `glob` takes so long in your project?

Let's implement your proposal then - update to the latest glob and stop following symlinks. While I am concerned that such solution may bite us in the future, it is improving the situation in the short term, thus it's worth the risk.

Just make sure the commit message clearly describes the breaking change it introduces.

jldec:
@bajtos - Yes, circular symlinks was the main cause of the long delays in my case.

I agree that given #370, bumping the glob version which disables symlink following is the right solution for now. I will post a separate PR for that.

Hopefully having an up-to-date glob version is another small incentive for some NI user to contribute fixes (a) to enable glob to return realpaths  and (b) to enable symlink following in glob (with cycle detection)

I also like your suggestion of using a glob pattern for the NI preload option

PR #551, opened 2015-02-06
use glob v4.3.5 - NOTE disables symlink following

jldec:
because of #370 symlinked directories don't work in NI.
this change at least prevents massive delays when there are circular symlinks. (like #498)

further discussion in #550

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,11 @@ Node Inspector supports almost all of the debugging features of DevTools, includ
 
 ## Known Issues
 
+* Automatic preloading of js files to set breakpoints will not work in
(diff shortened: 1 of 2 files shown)


bajtos:
@jldec Thank you for the new pull request. Could you please also include a short note in the README explaining the situation? "No" is an acceptable answer ;-)

jldec:
@bajtos 
README note added. hth

(b.t.w. i think the Travis CI failure is the same timeout you tried to mitigate before)

bajtos:
Landed, thank you for the contribution and especially the docs part. :bow:

bajtos:
Released as `node-inspector@0.9.0`.

Issue #534, opened 2015-01-23
Script is resumed as soon as node-inspector is loaded


ELLIOTTCABLE:
I'm launching my program with `--debug-brk`. Execution is immediately paused. However, as soon as I load up `node-inspector`'s page in Chrome, the script resumes, and immediately crashes (with the error I'm trying to debug). This gives me no time to set, say, a breakpoint. (Incidentally, `debugger;` statements don't help.)

---

**Affects node v0.11.14 and higher. Use an earlier version of node to unblock yourself**

 - v0.11.13 (Last compatible Unstable) - http://nodejs.org/dist/v0.11.13/
 - v0.10.38 (Last compatible Stable) - http://nodejs.org/dist/v0.10.38/

Node v0.10.38 will give you the profiler 'partial support' warning which you can ignore if you do not use the profiler. Use v0.11.13 as the warning suggests if you need the profiler.

---

I'm not sure what other information I can share, here. My own code is fairly complex; if you'd like to try and reproduce this on your end, the following should approximately duplicate the environment where I'm seeing this:

    brew install nvm
    nvm install 0.11
    nvm use 0.11
    npm install -g node-inspector
    node-inspector &
    git clone git@github.com:Paws/Rulebook.git --branch letter-01
    git clone git@github.com:ELLIOTTCABLE/Paws.js.git --branch support-rulebook
    cd Paws.js
    npm install
    npm install # yes, run it twice. bugs in my codebase. unrelated.
    node --debug-brk ./Executables/paws.js check "../Rulebook/The Letters/01. On rules, meaning, and consistency.rules.yaml"
    # At this point, it should be paused ...
    # ... but as soon as you load up http://127.0.0.1:8080/debug?port=5858, for me, it resumes and crashes.

3y3:
Please share also error message.

3y3:
Can you reproduce this behavior on some siple app? (for example [node-debug-samples](https://github.com/node-inspector/node-debug-samples))

ELLIOTTCABLE:
@3y3 from what? There's no error, it just opens, flashes the inspector for a moment, and then my code resumes-fails-exits, returning the inspector to the “Hit Cmd+O to open a file” view.

ELLIOTTCABLE:
re: simpler codebase, I'll try that right now. (Wasn't expecting this quick of a reply at 4AM Chicago time, heh; about to crash soon. FYI.)

ELLIOTTCABLE:
[...]

PR #560, opened 2015-02-15
More compatible with iojs

3y3:
We are not full compatible, but can use old features.

@bajtos , please review.

### DIFF
--- a/test/BreakEventHandler.js
+++ b/test/BreakEventHandler.js
@@ -1,0 +1,98 @@
+var expect = require('chai').expect,
+    launcher = require('./helpers/launcher'),
+    InjectorClient = require('../lib/InjectorClient').InjectorClient,
+    BreakEventHandler = require('../lib/BreakEventHandler').BreakEventHandler,
+    ScriptManager = require('../lib/ScriptManager').ScriptManager;
+
+describe('BreakEventHandler', function() {
+  var childProcess,
+      session;
+
+  describe('event cache', function() {
+    it('works correctly in debug-brk mode', function(done) {
+      prepareBreakEventHandler(true, function() {
+        expect(session.breakEventHandler._eventsCache).to.instanceof(Array);
+        expect(session.breakEventHandler._eventsCache.length).to.equal(0);
+
+        session.debuggerClient.request('continue', undefined, function() {
+          childProcess.stdin.write('pause\n');
+          setTimeout(function() {
+            expect(session.breakEventHandler._eventsCache.length).to.equal(1);
+            done();
+          }, 100);
+        });
+      });
+    });
+
+    it('works correctly in debug mode', function(done) {
+      prepareBreakEventHandler(false, function() {
+        expect(session.breakEventHandler._eventsCache).to.instanceof(Array);
+        expect(session.breakEventHandler._eventsCache.length).to.equal(0);
+
+        childProcess.stdin.write('pause\n');
+        setTimeout(function() {
+          expect(session.breakEventHandler._eventsCache.length).to.equal(1);
+          done();
+        }, 100);
+      });
+    });
+  });
+
+  describe('event resolver', function() {
+    before(function(done) {
+      prepareBreakEventHandler(false, function() {
+        childProcess.stdin.write('pause\n');
+        setTimeout(done, 100);
+      });
+    });
+
+    it('should resolve events on `resource-tree-resolved` event', function(done) {
+      // Test checks that the event was resolved from cache and triggered (sent to the client).
+      // It fails on timeout when the implementation is not correct. 
+      session.frontendClient.once('Debugger.paused', function() {
+        expect(session.breakEventHandler._eventsCache).to.equal(null);
+        done();
+      });
+
+      expect(session.breakEventHandler._eventsCache).to.be.instanceof(Array);
+      expect(session.breakEventHandler._eventsCache).to.have.length(1);
+
+      session.debuggerClient.request('scripts', null, function(error, result) {
+        if (error) done(error);
+
+        result.forEach(session.scriptManager.addScript.bind(session.scriptManager));
+        session.emit('resource-tree-resolved');
+      });
+    });
+
+    it('should resolve events immediately if resource tree was resolved', function(done) {
+      // This test is relying on the global state changes made by the previous test
+      // and checking that continue + pause will be correctly handled.
+      // It fails on timeout when the implementation is not correct. 
+      session.frontendClient.once('Debugger.paused', function() {
+        done();
+      });
+
+      session.debuggerClient.request('continue', null, function() {
+        childProcess.stdin.write('pause\n');
+      });
+    });
+  });
+
+  function prepareBreakEventHandler(pauseOnStart, callback) {
+    launcher.runCommandlet(pauseOnStart, function(_childProcess, _session) {
+      session = _session;
+      childProcess = _childProcess;
+
+      session.resourceTreeResolved = false;
+      session.once('resource-tree-resolved', function() {
+        session.resourceTreeResolved = true;
+      });
+      session.injectorClient = new InjectorClient({}, session);
+      session.scriptManager = new ScriptManager({}, session);
+      session.breakEventHandler = new BreakEventHandler({}, session);
+
+      callback();
+    });
+  }
+});
--- a/test/HeapProfilerAgent.js
+++ b/test/HeapProfilerAgent.js
@@ -120,11 +120,10 @@ function initializeProfiler(done) {
     heapProfilerAgent = new HeapProfilerAgent({}, session);
 
     injectorClient.once('inject', function(injected) {
-      if (injected) done();
+      if (injected) debuggerClient.request('continue', null, done);
     });
-    injectorClient.once('error', function(error) {
-      done(error);
-    });
+    injectorClient.once('error', done);
+
     injectorClient.inject();
   });
 }
--- a/test/HeapProfilerAgent.js
+++ b/test/HeapProfilerAgent.js
@@ -7,15 +7,8 @@ var expect = require('chai').expect,
 
 var heapProfilerAgent,
     heapProfilerClient,
-    debuggerClient;
-var frontendClient = new EventEmitter();
-frontendClient.sendEvent = function(event, message) {
-  this.emit(event, message);
-};
-frontendClient.sendLogToConsole = function(level, text) {
-  console[level](text);
-};
-frontendClient.off = frontendClient.removeListener;
+    debuggerClient,
+    frontendClient;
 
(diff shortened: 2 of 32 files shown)


3y3:
Sorry, not ready to review

bajtos:
> Sorry, not ready to review

No worries, please leave a comment when it's ready.

3y3:
@bajtos , this partially ready to review.

Strategy of this pr:
1. Refactor session module. Inherit it from EventEmitter.
2. Restructure all modules to receive `session` object instead of `debuggerClient, frontendClient etc.`
3. Fix test after refactoring
4. Fix test for 0.12 - to be sure that we work on 0.12 with new changes (there are some speculative changes. I need help to understand how they work)
5. Emit `resource-tree` event from session instead of PageAgent. Rename to `resource-tree-resolved`
6. Add new tests for `resource-tree-resolved` feature

About fixes for 0.12:
Ok, I can understand why we need to resume debugger after injections [like there](https://github.com/3y3/node-inspector/commit/c95736904bdf37987c4115710bbb381ac17937c0#diff-78792a5f176ac27c2f69d3b51e9c0b1bR200) - this is related to new isolated debugger with his own event loop.
But that happens [here](https://github.com/3y3/node-inspector/commit/c95736904bdf37987c4115710bbb381ac17937c0#diff-e26d3be456d1f6213df7d90212089743R165), how it's possible what `fn` is undefined?
I inspect this situation long time:
- If you try to pass all tests without this change - failure any time
- If you try to pass only tests for InjectorClient without this change - failure some time
- If you try to pass only failing test without this change - all works. WTF?!

It may be that we need help of core node developers here. Unfortunately I can't reproduce this problem in isolated sample.

test/BreakEventHandler.js
```diff
@@ -0,0 +1,93 @@
...
+    InjectorClient = require('../lib/InjectorClient').InjectorClient,
+    BreakEventHandler = require('../lib/BreakEventHandler').BreakEventHandler,
+    ScriptManager = require('../lib/ScriptManager').ScriptManager;
+    
+describe('BreakEventHandler', function() {
+  var childProcess,
+      session;
+
+  describe('should cache events', function() {
+    it('if started in debug-brk mode', function(done) {
```
bajtos: The test names "describe should change events" and "it if started in debug-brk mode" don't read as English. Can you please rephrase? For example:

```js
describe('event cache', function() { // or 'event caching'
  it('works correctly when started in debug-brk mode', function(done) {
```

The same applies to all other test names.

test/BreakEventHandler.js
```diff
@@ -0,0 +1,93 @@
...
+        childProcess.stdin.write('pause\n');
+        setTimeout(function() {
+          expect(session.breakEventHandler._eventsCache.length).to.equal(1);
+          done();
+        }, 100);
+      });
+    });
+  });
+  
+  describe('should resolve events', function() {
```
bajtos: Perhaps `describe('event resolver')`?

test/BreakEventHandler.js
```diff
@@ -0,0 +1,93 @@
...
+      });
+      
+      session.debuggerClient.request('scripts', null, function(error, result) {
+        if (error) done(error);
+
+        result.forEach(session.scriptManager.addScript.bind(session.scriptManager));
+        session.resourceTreeResolved = true;
+        session.emit('resource-tree-resolved');
+      });
+    });
```
bajtos: What is the purpose of this test? I find it very difficult to understand. I don't see any asserts (expectations), are you testing only that the debugger does not crash?

test/BreakEventHandler.js
```diff
@@ -0,0 +1,93 @@
...
+    it('immediately after resource-tree-resolved', function(done) {
+      session.frontendClient.once('Debugger.paused', function() {
+        done();
+      });
+
+      session.debuggerClient.request('continue', null, function() {
+        childProcess.stdin.write('pause\n');
+      });
+    });
+  });
```
bajtos: What is the purpose of this test? I don't see any asserts (expectations), are you testing only that the debugger does not crash?

test/BreakEventHandler.js
```diff
@@ -0,0 +1,93 @@
...
+      session.resourceTreeResolved = false;
+      
+      var injectorClient = new InjectorClient({}, session);
+      session.injectorClient = injectorClient;
+      
+      var scriptManager = new ScriptManager({}, session);
+      session.scriptManager = scriptManager;
+      
+      var breakEventHandler = new BreakEventHandler({}, session);
+      session.breakEventHandler = breakEventHandler;
```
bajtos: Can you use `SessionStub` here?

test/HeapProfilerAgent.js
```diff
@@ -188,23 +184,23 @@ describe('HeapProfilerClient', function() {
...
+    debuggerClient = session.debuggerClient;
+    frontendClient = session.frontendClient;
+
+    var injectorClient = new InjectorClient({}, session);
+    session.injectorClient = injectorClient;
+
+    heapProfilerClient = new HeapProfilerClient({}, session);
+    session.heapProfilerClient = heapProfilerClient;
+
+    heapProfilerAgent = new HeapProfilerAgent({}, session);
```
bajtos: Is there any reason why this initialisation is not done automatically by `SessionStub`?

bajtos:
I like the idea of moving the different object instances to a single session object. My initial dependency-injection approach did not worked well. Can you please make `SessionStub` more feature complete, so that it can be used as a drop-in replacement for a real `Session` object in as many tests as possible?

The changes made in the production code (`lib/*`) look good to me.

> About fixes for 0.12:
Ok, I can understand why we need to resume debugger after injections like there - this is related to new isolated debugger with his own event loop.
But that happens here, how it's possible what fn is undefined?

Perhaps `debug-brk` stops the process before `console.log` is defined?

test/BreakEventHandler.js
3y3: This is a main test for #534, so I test what we have consistent state and `BreakEventHandler` 
doesn't ignore events [here](https://github.com/3y3/node-inspector/commit/27dd9cabd239c60c748d77d9d1f27adacec47787#diff-3cf7bfb8d4bd4a7cd3212e0854ee8523R96)

```
describe('event resolver', function() {
  it('should send cached events to frontend on `resource-tree-resolved`');
  it('should send events to frontend after `resource-tree-resolved`');
});
```

It meybe reasonable to check `eventsCache.length` here.

test/HeapProfilerAgent.js
3y3: Do you mean, that module should register itself?

```js
function HeapProfilerAgent(config, session) {
  ...
  session.heapProfilerAgent = this;
  ...
}
```

3y3:
> Can you please make SessionStub more feature complete, so that it can be used as a drop-in replacement for a real Session object in as many tests as possible

Can you, please, describe more verbose this part?

test/HeapProfilerAgent.js
3y3: Or somethink like this
```
session.init('InjectorClient', 'HeapProfilerAgent', 'HeapProfilerClient');
```

test/BreakEventHandler.js
bajtos: So what is the failure message of this test when your fix is not applied?

#534 says

>  it seems that NodeJS 0.12 now signals the initial break as an actual breakpoint, and if you treat that as an unknown breakpoint and just instruct the debugger to continue, then the app immediately resumes and runs to completion.

> If I'm reading this correctly, their fix is to treat unknown breakpoints differently (basically; not instruct the debugger to continue on) while the debugger hasn't finished attaching itself.

Is there a way how to (re)write this unit-test to verify that the process was paused instead of running to completion? Or do I understand this wrong and the tests checks only BreakEvenHandler behaviour in isolation?

> BreakEventHandler  doesn't ignore events here

In that case you should be able to check that the event was triggered (sent to the client), is that correct? Is that what `session.frontendClient.once('Debugger.paused'` is doing?

Maybe it's starting to make more sense now. So your test is failing on a timeout before the fix is applied, because all breakpoints are ignored and the program will be never paused.  With the fix in place, the tests passed via `done` call from the event listener. Is that a correct description? Can you please add a comment to the test explaining this?

test/BreakEventHandler.js
bajtos: IIUC, this test is relying on the global state changes made by the previous test and checking that continue + pause will be correctly handled. It fails on timeout when the implementation is not correct. If that's the case, then again add an explanatory comment please.

test/HeapProfilerAgent.js
bajtos: If you look at `lib/session.js`, you can see that the session object provides these objects:

```
+  this.debuggerClient = new DebuggerClient(debuggerPort);
+  this.frontendClient = new FrontendClient(wsConnection);
+  this.injectorClient = new InjectorClient(config, this);
+  this.consoleClient = new ConsoleClient(config, this);
+  this.heapProfilerClient = new HeapProfilerClient(config, this);
+  this.scriptManager = new ScriptManager(config, this);
+  this.breakEventHandler = new BreakEventHandler(config, this);
+  this.frontendCommandHandler = new FrontendCommandHandler(config, this);
```

I am saying that the `SessionStub` used in tests should provide all of these objects too. So that you don't have to create e.g. `heapProfilerClient` manually (like in the test here), but instead you can rely on the fact that `SessionStub` provides the same set of objects as the real `Session`.

bajtos:
> > Can you please make SessionStub more feature complete, so that it can be used as a drop-in replacement for a real Session object in as many tests as possible

> Can you, please, describe more verbose this part?

Sure, see https://github.com/node-inspector/node-inspector/pull/560#discussion_r28327413

test/HeapProfilerAgent.js
3y3: But in this way all our tests will be dependent on all our modules.
If I test `ConsoleAgent`, I don't want to initialize `HeapProfilerAgent`, but this will happens, because I initialise `SessionStub`, what will contain `HeapProfilerAgent` initialization.
Or I'm wrong here?

bajtos: I see, that makes sense. Ignore my comment then.

3y3:
Added description for tests. Rebased to master.
Ready to review.

3y3:
I'll change names on next fix step.
Waiting resolution about `FrontendCommandHandler` test.

bajtos:
No more comments. I'd really like to see `waitResourceTree` renamed because it does not follow English grammar IMHO, but I can live with the current version too.

3y3:
@bajtos , I'm really sorry for my English =) You feel free to correct me any time.

3y3:
Ok. I'll make last fixes and merge this.

PR #564, opened 2015-02-19
.travis.yml: drop 0.11, add 0.12 and io.js

bajtos:
This PR will most likely fail the build, we have two options: either land it to make it clear that Node Inspector is not supported on these versions yet, or keep the PR open until support for io.js & Node v0.12 is landed.

@3y3 Thoughts?


bajtos:
Cool, landed.

PR #575, opened 2015-03-01
another way of starting a node-inspector debug session in windows

hansrwindhoff:


bajtos:
@3y3 you are running on Windows, could you please review?

hansrwindhoff:
Here is a screen shot of the four involved windows: 
- the command prompt that starts the three processes, 
- the iojs or node window running the app, 
- the node-inspector (here running in iojs)
- and the chrome dev tools sitting on a break point and showing some info in the wathc window
image

this is under windows 8.1 64bit

hansrwindhoff:
... the same windows running in node (0.12.0):
image

README.md
```diff
@@ -268,6 +268,16 @@ an undocumented API function `process._debugProcess(pid)`:
...
     ```
 
+Another way is to start these three processes:
+
+
+    start "" /D"." node --debug-brk app.js
+    start "" /D"." node-inspector &
+    start chrome http://127.0.0.1:8080/debug?port=5858
+
+You can call these (each line is a command) from the command prompt or put all of them into a batch/or cmd file and then run that. In this example app.js is the applications entry point.
```
1j01: Missing apostrophe: "application's"

Issue #578, opened 2015-03-06
My vision of next node-inspector startup system


3y3:
This is very subjective description of some features related with `node-inspector` startup, what I'd like to see in next versions.
I'm not sure that all of this is realisable, or described correctly.

### `ni` shortcut
To start node-inspector use
```bash
$ ni
```

### Documented `--config` option
```bash
$ ni --config ~.ni-mocharc
$ ni --config ~.ni-gulprc
```
#### Multy config sources
```bash
$ ni --config ~.ni-ignorerc --config ~.ni-nodeflagsrc
```

#### `debug` subcommand
Like `node-debug` command
```bash
$ ni debug app.js
$ ni debug --config ~.ni-gulprc app.js
```

#### `config` subcommand
Like `git config`, but with some special features.
```bash
$ ni config --global web-host 127.0.0.2
$ ni config --global --unset web-host
$ ni config [--local] web-host 127.0.0.2
$ ni config --name ~.ni-gulprc web-host 127.0.0.2
```
So, will be useful to search `.node-inspectorrc` in app folder (Or maybe `.nirc`. I'd like `.nirc`:+1:)

#### `url` subcommand
```bash
# http://127.0.0.1:8080/debug?port=5858&ws=127.0.0.1:8080
$ ni url
# http://127.0.0.1:8081/debug?port=5858&ws=127.0.0.1:8081
$ ni url --web-port 8081

$ ni url | chrome
```

#### `install` command
For future extensions system.
It's not simple alias for npm install. We well need to integrate subpackages to `node-inspector` structure
Like `sln install`
```bash
$ ni install mongodb
$ ni install mongodb-mongose
$ ni install network-ws
$ ni install console-ascii-formatting
```

auchenberg:
I like the short ``ni`` keyword, and the overall simpler grouping of commands. 

:+1:

marcominetti:
:+1: Please just mind about programmatic node-inspector startup. The implementation of the next startup system could be implemented through a exportable interface to load node-inspector as a module.

var ni = require('node-inspector');
ni.debug([target][,options]); < in case of missing target the ni could be started within the current process
ni.config = f(options[,global:false][,cb]);
ni.url = f([options][,cb]);
ni.install(modules[,options][,cb]) < similar to npm

The return values could be Promise in case of missing cb, and sync implementation could be separated the node way (exec,execSync).

I currently launch node-inspector spawning process (maybe good for exception isolation?).

[...]

Issue #502, opened 2014-12-20
support ansi console output


nikmartin:
Using node-inspector with a debug module like debug makes debug in the node-inspector console output look like:
`[96mapp [36m[90mHEAP USAGE: {"num_full_gc":8,"num_inc_gc":5,"heap_compactions":8,"usage_trend":0,"estimated_... (length: 149)[36m +3s[0m`

vs:

`app HEAP USAGE: {"num_full_gc":8,"num_inc_gc":5,"heap_compactions":8,"usage_trend":0,"estimated_base":19313048,"current_base":19313048,"min":16572648,"max":19313048} +3s`

ANSI support isn't necessary, but maybe eat non ASCII charatcers?
##

bajtos:
Cross posting my comment from #595:

> AFAIK, [debug](https://www.npmjs.com/package/debug) supports color output in browser console too, it may be a good source of inspiration.

IMHO, Node Inspector should provide at least very basic ANSI support out of the box. Perhaps we can discard all ANSI control codes by default, and then allow a user plugin to override this behaviour? But then converting ANSI control codes to HTML is a solved problem (see [ansi-to-html](https://www.npmjs.com/package/ansi-to-html)), thus it may be pretty easy to implement full support too.

mgcrea:
:+1: I'm using `bunyan`+`chalk` for nice console colored output, and log messages are unfortunately barely readable with node-inspector. I do agree with @bajtos, Striping ANSI control codes would be at least a good start (eg. drop coloring output).

Issue #180, opened 2013-07-26
Add support for extensions (plugins)


pmuellr:
Is there any way to provide extensions?  From memory, there's enough extension goop in web inspector that something could be cobbled together.  It doesn't need to work like Chrome - no need to only support inspector extensions only as Chrome extensions - I should be able to add an extension as a node package, I guess.

I'd like to add some profiling, like [nodprof](https://github.com/pmuellr/nodprof).  And a module explorer.
##

marcominetti:
Ok guys, as soon as the new frontend #492 is merged, I'll start working on the plugin system as described by @3y3 in #582. +1

nascarjake:
I'm looking to extend node-inspector by building my own DevTool extensions. Are we any closer to this? It looks like since 3 years ago the Profiles tab found it's way in.

PR #582, opened 2015-03-09
Added ECMAScript 6/7 support to console

marcominetti:
Hi there,
I added ECMAScript 6/7 experimental support through https://github.com/babel/babel project (@sebmck).

To enable it, append &babel=7 to the node-inspector URL. The babel param value can be 5, 6 or 7 to enable specific ECMAScript version.

The code is transpiled only if babel param is >5 and only whether ECMAScript 6 or 7 code is detected. ECMAScript detection is carried out through acorn-babel parsing exceptions.

It requires to manually load babel/polyfill with require('babel/polyfill').

TODO: automatic build/loading from latest babel and acorn-babel, proper handling of Promise returned values (resolve to console/reject to error), automatically inject babel/polyfill, write documentation, choose to migrate transformation back to the backend (prior versions)


marcominetti:
Just to try it.

test.js file:
```javascript
require('babel/polyfill')
require('http').createServer(function(req,res){
  res.end('hello world')
}).listen(8081);
```

Environment start:
```bash
npm install babel
node --debug test.js
```

Launch node-inspector and open http://127.0.0.1:8080/debug?port=5858&babel=7

Execute in the node-inspector console:
```javascript
global.sleep = function (ms) {
  return new Promise(function (resolve, reject) {
    setTimeout(function () {
      resolve({});
    }, ms)
  });
};
```

Then try:
```javascript
await sleep(2000); console.log('hi');
```

To avoid console errors just pause execution through debugger, but await (and promises) will be paused too.

3y3:
Hello, @marcominetti . I'm not sure that making `babel` a part of `node-inspector` is a right thing.
But this is a good point to start extended discussion about plugins.
For example, we have [one other](https://github.com/node-inspector/node-inspector/issues/502) feature issue, about highlighting `debug` module colors in console, but this is useful only for users that use `debug` module in their projects, so this can't be implemented as main part of node-inspector.

I'd like to provide to users simple way to extend `ni` by plugins.
This is a work on two-three fronts:
1. frontend extension
2. inspector backend extension
3. [app extension]

Your current pr takes only first part. So, how I see frontend extension mechanism:
1. We can store all plugins in subdirectory with same name. Example:
```
node-inspector
-> plugins
->-> babel
```
2. Each plugin has name in npm `node-inspector-<plugin name>`. Example `node-inspector-babel`
3. Each plugin has `module.json`, like [this](https://github.com/3y3/node-inspector/blob/frontend/front-end-node/module.json), [inspector.json](https://github.com/3y3/node-inspector/blob/frontend/front-end-node/inspector.json) and optionally [protocol.json](https://github.com/3y3/node-inspector/blob/frontend/front-end-node/protocol.json). We need to compile extension structure on node-inspector startup, or for each new [ws session](https://github.com/node-inspector/node-inspector/blob/master/lib/debug-server.js#L20-L23)
4. Add new paths to [express router](https://github.com/3y3/node-inspector/blob/frontend/lib/debug-server.js#L61-L64), like
```js
app.use('/plugins', express.static(path.join(__dirname, '../plugins')));
app.get('/inspector.json', function(req, res) {
  res.send(compiled_in_memory_inspectorjson);
});
```
Where `compiled_in_memory_inspectorjson` is the result of something like this:
```js
function compileInspectorJson() {
  var base = require('front-end-node/inspector.json');
  var plugins = fs.readdirSync('plugins');
  plugins.forEach(function(pluginName) {
    var ext = require(pluginName + '/inspector.json');
    base.push.apply(base, ext);
  });
  return JSON.stringify(base);
}
```


So for example if we have `babel` plugin:
`cd node-inspector/plugins/babel`
`echo module.json`
```js
{
    "dependencies": [ ],
    "scripts": [
        "BabelPlugin.js"
    ]
}
```
`BabelPlugin.js` is your current diff for overrides.

`echo inspector.json`
```js
[
  { "name": "babel", "type": "autostart" },
  { "name": "babel/subplugin1", "type": "autostart" },
  { "name": "babel/subplugin2" }
]
```
`echo protocol.json`
```js
{
    "domains": [
      {
        "domain": "Babel",
        "types": [],
        "commands": [],
        "events": [
            {
                "name": "testBabelEvent",
                "description": "Only for test"
            }
        ]
      }
    ]
}
```

And `compiled_in_memory_inspectorjson` is equal to:
```
[
  { "name": "node", "type": "autostart" },
  { "name": "node/console", "type": "autostart" },
  { "name": "node/sources", "type": "autostart" },
  { "name": "node/settings", "type": "autostart" },
  { "name": "node/main", "type": "autostart" },
  { "name": "audits", "type": "exclude"},
  { "name": "network", "type": "exclude"},
  { "name": "elements", "type": "exclude"},
  { "name": "timeline", "type": "exclude"},
  { "name": "resources", "type": "exclude"},
  { "name": "babel", "type": "autostart" },
  { "name": "babel/subplugin1", "type": "autostart" },
  { "name": "babel/subplugin2" }
]
```

We need to compile same way `protocol.json`.


So, @marcominetti , are you ready to implement plugin system? This work is relative to `frontend` pr. Unfortunately in current time I work on other `ni` issues, and can't finish `frontend` pr at soon, but I'm ready to review prs relative to it.
So keep in mind that this is a discussion about plugins, and your thoughts are important.
Welcome to discussion on [gitter](https://gitter.im/node-inspector/node-inspector)

/cc @bajtos

metaford:
NAVER - http://www.naver.com/
--------------------------------------------

z.lee@navercorp.com 님께 보내신 메일 <Re: [node-inspector] Added ECMAScript 6/7 support to console (#582)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

3y3:
@metaford , shut up fucking bot.

bajtos:
> @metaford , shut up fucking bot.

@3y3 you can report an abuse using this link: https://github.com/contact?report=metaford I already did that myself.

3y3:
> @3y3 you can report an abuse
Already reported

marcominetti:
Interesting, I need to review your frontend branch to get some thoughts... btw, why are you defining npm naming convention? Is it related to your next startup system proposal (ni install)? Or... are u thinking about a "build" phase that merges code from /node_modules/node-inspector-xxx folders into /node-inspector/plugins/?

marcominetti:
Ok @3y3, with the node-inspector-<plugin_name> convention you meant defining naming for the custom ni distribution with embedded plugins, did you? Ok, I can do develop the plugin system. Do you want to create a Feature request issue or can continue here?

3y3:
> Ok, I can do develop the plugin system.

@marcominetti , good news!

Feature request already exists (#180). I'll update information.

> why are you defining npm naming convention?

> Is it related to your next startup system proposal (ni install)?

Yes. But it's out of scope of current feature. 
Current `install` solution - clone repo into plugins folder =)

bajtos:
> Yes. But it's out of scope of current feature. 
> Current install solution - clone repo into plugins folder =)

IMO that pretty much sucks, as users have to find the place where Node Inspector is installed and they also can't use tagged releases. We really need to find a better solution before we can call plugin implementation as done.

3y3:
> Current install solution - clone repo into plugins folder =)
> > IMO that pretty much sucks

@bajtos , I'm agree. Please read my thoughts in `install` part of #578

marcominetti:
@3y3 @bajtos ...a solution could be integrate npm (through npm link instead of having it in ni deps) and just look for every node module - installed locally (cwd) and globally - that matches node-inspector-plugin-xxx name.

IMO the naming convention could be:
  * node-inspector-xxx: custom inspector bundled with plugins (out of scope for the plugin system feature
  * node-inspector-plugin-xxx: plugin that extends ni and contains the module.json, protocol.json and inspector.json and related files

At load time, ni will find the matching node-inspector-plugin-xxx modules, read jsons and create proper forwarders for frontend files. In this case I would suggest to define a dedicated frontend/backend folders within the xxx plugin to avoid security concers of serving backend files to the frontend.

To sum up:
```
node-inspector
-> plugins
->-> babel
->->-> module.json
->->-> protocol.json
->->-> inspector.json
->->-> front-end    < only this will be served through express with URI prefixed with /babel/
->->-> lib
```
The same for npm installed modules:
```
[cwd]
-> node_modules/
->-> node-inspector-plugin-babel
->->-> module.json
->->-> protocol.json
->->-> inspector.json
->->-> front-end    < only this will be served through express with URI prefixed with /babel/
->->-> lib
```

In case of naming conflicts (same plugin embedded in ni /plugins folder and available as npminstalled pkg), the ni embedded wins (cause npm updated could break currently installed ni).

In module.json could be useful to add - by specification - a property for supported ni version(s). This property could be checked with semver during load phase.

Any thoughts?

bajtos:
> node-inspector-plugin-xxx

That's a super long name. How about `ni-xxx` and using npm keywords like "node-inspector" + "plugin" for discoverability?

marcominetti:
@bajtos mmm... is it really reliable to use keywords for discoverability? I mean composition of node-inspector and plugin could be not directly related to the plugin system of node-inspector...

What about

> ni-plugin-xxx

Could be shorter then loopback-connector-xxx... ;)

Is node-inspector already published as "ni"?

bajtos:
Use keywors for discoverability on npmjs.org.

The way how node-inspector discovers plugins should be based entirely on module name patter, either `ni-xxx` or `ni-plugin-xxx`. 

> Is node-inspector already published as "ni"?

No, and I don't think it will ever be.

marcominetti:
Ok, got it. Next monday I'll start working on the plugin system. :+1:

PR #586, opened 2015-03-13
use glob v5.0.3 to follow symlinks and return realpaths

### LINKED ISSUE #370: Cannot set breakpoints in module loaded with npm link

jldec:
The glob module now has options to return realpaths and follow symlinks. With glob v5.0.3. this feature is usable with node-inspector.

Preloaded scripts in symlinked directories now show up properly in the inspector source tree and breakpoints work, fixing issue #370.

f.w.i.w. i also added a little debug tracing in ScriptFileStorage. e.g. to trace during tests do:

    DEBUG=node-inspector:ScriptFileStorage npm test

Fix #370

### DIFF
--- a/lib/ScriptFileStorage.js
+++ b/lib/ScriptFileStorage.js
@@ -2,6 +2,7 @@ var fs = require('fs');
 var path = require('path');
 var async = require('async');
 var glob = require('glob');
+var debug = require('debug')('node-inspector:ScriptFileStorage');
 
 var MODULE_HEADER = '(function (exports, require, module, __filename, __dirname) { ';
 var MODULE_TRAILER = '\n});';
@@ -149,17 +150,20 @@ $class.listScripts = function(rootFolder, pattern, callback) {
   //    callback
   // );
 
+  debug('glob %s on %s', pattern, rootFolder);
+
   glob(
     pattern,
     {
       cwd: rootFolder,
+      follow: true,
+      realpath: true,
       strict: false
     },
     function(err, result) {
       if (result) {
-        result = result.map(function(relativeUnixPath) {
-          var relativePath = relativeUnixPath.split('/').join(path.sep);
-          return path.join(rootFolder, relativePath);
+        result = result.map(function(unixPath) {
+          return unixPath.split('/').join(path.sep);
         });
       }
 
@@ -170,6 +174,7 @@ $class.listScripts = function(rootFolder, pattern, callback) {
         result: result
       };
 
+      debug('glob returned %s files', err || result.length);
       callback(err, result);
     }.bind(this)
   );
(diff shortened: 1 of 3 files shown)


lib/ScriptFileStorage.js
```diff
@@ -149,17 +150,20 @@ $class.listScripts = function(rootFolder, pattern, callback) {
...
+      realpath: true,
       strict: false
     },
     function(err, result) {
       if (result) {
-        result = result.map(function(relativeUnixPath) {
-          var relativePath = relativeUnixPath.split('/').join(path.sep);
-          return path.join(rootFolder, relativePath);
+        result = result.map(function(unixPath) {
+          return unixPath.split('/').join(path.sep);
```
bajtos: Are you sure the paths are returned as absolute, even though you specify rootFolder via `options.cwd` and not as `options.root`?

Related: 1c06f8ae6 and https://github.com/isaacs/node-glob/commit/f3105a3cfc12d3c6e0a2c931d8cf7f1a6a1ff49a

bajtos:
Thank you for the pull request, sorry for taking so long to review it.

Can you please rebase it on top of the current master and address my comment above?

lib/ScriptFileStorage.js
jldec: yes - i had to fix the code above to avoid prepending the rootfolder twice (which breaks inspector). i also had a little exchange with isaacs about the realpath returned paths being absolute now (and the match results not). https://github.com/isaacs/node-glob/issues/184

jldec:
rebased and comment addressed.
I also moved to the latest glob version which has updated absolute-path polyfill.
https://github.com/isaacs/node-glob/commit/855230118851fb7d4a6cd7616851f03bca12ad06
Hopefully this will improve things under windows. (i don't have a test setup for windows)

bajtos:
Landed via 5e368e8. I have improved the README text a bit, squashed the commits and added more details to the commit message.

Thank you for the contribution!

PR #610, opened 2015-04-13
Add note regarding breaking bug

ELLIOTTCABLE:
As requested by @bajtos and noted by @dpwolfe in #534.


dpwolfe:
Thanks for creating this PR.  LGTM

bajtos:
Landed via 24e83f3, I have made few improvements:

 - moved the warning below the icons
 - reformatted it to a regular text instead of a reply
 - wrapped all lines at 80 characters

Thank you for the contribution!

PR #616, opened 2015-04-17
Updated frontend: fixed debug port not read from URI.

marcominetti:


front-end-node/NodeInspectorOverrides.js
```diff
@@ -58,6 +60,18 @@ WebInspector.NodeInspectorOverrides.prototype = {
       showMainAppFile,
       null
     );
+  },
+  
+  _mergeConnectionQueryParams: function(){
+    var RuntimeQueryParamsObject = Runtime._queryParamsObject;
+    if (RuntimeQueryParamsObject['ws'] && RuntimeQueryParamsObject['port']){
+      if (RuntimeQueryParamsObject['ws'].indexOf('?') === -1){
```
3y3: Is there a situation then `ws` param contains `?`

3y3:
Please fix brackets style (missed spaces).
We use next syntax in node-inspector:
```js
if (statement) {

}

function(args) {

}
```

I don't see the reason to use `RuntimeQueryParamsObject ` (one long name instead of one other).
It may be reasonable to use short and readable `params`.

marcominetti:
Ok ;)

front-end-node/NodeInspectorOverrides.js
marcominetti: I don't know but some plugins could extend it, right? I would add it just to be safe... ;)

3y3:
Looks better. I'm ready to merge this one. Please squash changes.

marcominetti:
squashed

3y3:
Landed as 88921eb8d097df9e20864fa5b0a38c0e0bd7bbc6

PR #618, opened 2015-04-17
Updated frontend: added browser redirect from "/" root to proper inspector URI.

marcominetti:


bajtos:
I am proposing to drop `/debug` path from the URL and server the debugger UI directly on the root URL.

```diff
- http://localhost:8080/debug?ws=localhost:8080&port=5858
+ http://localhost:8080/?ws=localhost:8080&port=5858
```

@3y3 @auchenberg What's your opinion? Are you aware of any reason why the root URL may not work?

3y3:
I don't see any reasons to continue work on `/debug` path.
+1 to drop it (with backward compatibility).

marcominetti:
If @auchenberg agrees, I change this pr to drop it

bajtos:
@marcominetti let's not wait for @auchenberg and move this patch forward. Could you please rework the code to serve the debugger at the root url, as discussed above?

marcominetti:
@bajtos Yep, I should have time this afternoon or evening to carry out all ni related tasks... (CEST)

auchenberg:
Looks good. I don't have any problem with removing /debug, as long it won't
cause problems with /json or future endpoints.

bajtos:
> As long it won't cause problems with /json or future endpoints.

I think there should not be any problems. The (new) handler is supposed to serve only `GET /`, everything else should remain unchanged.

marcominetti:
done @bajtos

bajtos:
Landed via 21dee3d, I have modified the patch to update README too. Thank you for the contribution!

Issue #580, opened 2015-03-09
Inline source maps with .es6 file extension


jfaissolle:
It seems that when the filename extension is not .es6, inline source maps are not taken into account. I have two files:
test.js
```javascript
'use strict';
require('babel/register');
var test = require('./test2');
test();
```
and 
test2.es6
```javascript
'use strict';

module.exports = () => {
  let name = 'world';
  console.log(`Hello ${name}`);
};
```
`test2.es6` produces an inline source map but it is not used by `node-inspector`. If I rename it to `test2.js`, the source map works as expected.

bajtos:
I am afraid I don't have enough time to investigate this issue in detail,I suspect the problem may be related to what I have found in coffee-script transpiler - see https://github.com/node-inspector/node-inspector/issues/224#issuecomment-38785891. Here is the behaviour I would expect:

 - The transpiled javascript source has the source map embedded via data url (per rfc2397, see also strong-data-uri). No test2.map file is created.
 - The transpiled source is executed, the script file name is set to test2.js (must be a different file name!)
- The source map points to test2.es6 as the original source.

Possibly related: https://github.com/node-inspector/node-inspector/issues/100#issuecomment-31475540 and https://github.com/babel/babel/issues/306

3y3:
Fixed in 0.10.2.
Feel free to reopen this issue, if I'm wrong here.

Issue #566, opened 2015-02-23
Use Notification API on hitting breakpoints


wdbacker:
A very handy feature would be that during debugging in Chrome, the browser automatically can activate the tab where node-inspector is running when e.g. a breakpoint is encountered in the code.

I found something to do this using a chromium extension:
https://groups.google.com/a/chromium.org/forum/#!topic/chromium-extensions/x36zW3QsF0g

Or from the docs in the Chrome tabs API:
https://developer.chrome.com/extensions/tabs#method-update

Do you think this would be possible to implement in node-inspector without a lot of coding?

bajtos:
Hi @wdbacker, I am not familiar with Chrome extensions and tabs API enough to provide any meaningful advice. I am willing to accept this feature if somebody contributes the implementation though.

wdbacker:
Hi @bajtos, I just tried out some possible solutions for this feature. I came across a very obvious solution: if the node-inspector page encounters an active breakpoint, the node-inspector's javascript code can just do an alert('breakpoint encountered'); in the javascript code, and the Chrome browser will automatically switch to the node-inspector tab!

Btw, it would be handy if you can enable/disable this alert with a config option e.g. in the settings.

bajtos:
> If the node-inspector page encounters an active breakpoint, the node-inspector's javascript code can just do an alert('breakpoint encountered'); in the javascript code, and the Chrome browser will automatically switch to the node-inspector tab!

But then you have to close the alert window before you can continue your work, right? In my opinion, this would be very annoying, especially if the alert is raised even while the debugger tab is active.

wdbacker:
Yes, you'll have to press "Enter" on the alert once. But you're right, we need one more thing - checking if the debugging window is hidden:

if (document.visibilityState && (document.visibilityState === 'hidden'))
  alert('breakpoint encountered');

Now the alert is triggered only once and will not be annoying anymore.

Btw, the visibility API is supported starting from Chrome 13:
https://developers.google.com/chrome/whitepapers/pagevisibility

bajtos:
ok, the proposal looks reasonable now.

@3y3 what do you think?

3y3:
Be carefull - this realisation stuns websocket connection.
At first glance on current stage of our debugging protocol this will work safely.

The other thing  - this needs to be a configurable behavior.

wdbacker:
[...]

PR #621, opened 2015-04-17
Updated frontend: fixed search through Drawer panel.

marcominetti:
Implemented DebuggerAgent.searchInContent.
Implemented PageAgent.searchInResource.
Added /lib/search.js to provide shared search functions.
Added resource/script source caching.
Fixed highlighting matches bug on minified-file lines.


marcominetti:
@3y3 This is not an easy pick pr... I'll push some more commits with syntax and code cleanup for linter... but I know we'll talk about the source code caching in DebuggerAgent implementation... :D

3y3:
@marcominetti , sorry for pending of your features, but there is a lot of bugs, which I need to fix and release before review.

marcominetti:
@3y3, wow I've seen the new issues right now... may I help you? Feel free to assign me something. ;)

3y3:
Hello, @marcominetti , I target you from Plugin system pr. Also you can look on [0.10.2 milestone](https://github.com/node-inspector/node-inspector/milestones/0.10.2), here are bugs and features with hight priority. It is not necessary, what all this bugs needs to be fixed, in some situations we need only to describe solution. For example #566 - you are friendly with new front-end, so you can target man to entry point for this feature (or implement it by yourself).
So look at list, and if you ready to fix something - notify me.

marcominetti:
@3y3 I can work on #566 and #580...

marcominetti:
Let me know if you have any priority... ;)

marcominetti:
Please notify me when you will be ready for review this pr... cause I'm thinking about reworking it after changing the internal behaviour incase of --no-preload param...

3y3:
> I can work on #566 and #580...

Nice! For #580 look also on #628. I think it's a duplication. In #628 posted good example for tests. (For correct installation delete `npm-shrinkwrap`)

> Please notify me when you will be ready for review this pr... cause I'm thinking about reworking it after changing the internal behaviour incase of --no-preload param...

`--no-preload` was changed somewhere?

marcominetti:
> --no-preload was changed somewhere?

Not yet, I will elaborate the clue later on reviewing/reworking this pr... btw the ni stack is built to run "locally", so the --no-preload does prevent backend to feed frontend with every available file from disk... (am I right?) I use it to have frontend load quickly and smoothly in large project...

The problem is that the --no-preload flag also avoids frontend feeding the static content provider (at front-end) for actually loaded files... This dramatically impact search, cause at first run, it needs to fetch source code for every loaded file... I will try to override the --no-preload behaviour assuring that script content is fetched automatically for loaded files and stored within the static content provider...

cr0cK:
+1 for https://github.com/node-inspector/node-inspector/issues/628 !

a7madgamal:
Guys please work on this its a really important feature!
Thanks a million in advance :)

marcominetti:
hi @a7madgamal @3y3  I've successfully implemented the feature but there is a huge memory consumption cause I need to cache/preempt source files in arrays because search feature generates very huge number of getScriptSourceById/getResourceContent requests to the debugger through DebuggerClient and everything crashes... I could cache source in /tmp/ filesystem (but it's not so clever in cloud environments for the I/O)...

Solutions:
  1. fix debugger client or debugger backend crash on huge requests (best)
  2. implement an ordered queue at getScriptSourceById and getResourceContent (@3y3?)
  3. keep this implementation and/or move cache on disk (worst)

You can try implementation with ljve-inspector module available on npm (I use it actually).

a7madgamal:
I can't find it, could you provide the npm link?

marcominetti:
https://www.npmjs.com/package/ljve-inspector

3y3:
@marcominetti , I'm not familiar with search API in DevTools, but if I understand right main problem - we receive lot of parallel requests like `searchInContent` and we try to require sources for each `scriptId` also in parallel mode. So, we need to review strategy of original DevTools, when there exists a linked local folder:
- If in this case DevTools tries to search in all files (not only in loaded to front-end), then we need to completely avoid `getScriptSourceById` and use `fs` (this makes some problems for generated scripts, but I prefer to ignore this problems in initial implementation)
- Otherwise we can move search logic to app (look at injection example in `ConsoleAgent`)

In all this cases we need to completely avoid caching.

So, main question on current step - are we need to search in preloaded resources?

front-end-node/search/SearchOverrides.js
```diff
@@ -0,0 +1,121 @@
...
+  var createSearchRegex = function(query, caseSensitive, isRegex)
+  {
+    var regexFlags = caseSensitive ? "g" : "gi";
+    var regexObject;
+
+    if (isRegex) {
+      try {
+        regexObject = new RegExp('^.*?'+query+'.*?$|^.*?'+query+'.*?\n|\n.*?'+query+'.*?\n|\n.*?'+query+'.*?$', regexFlags);
+      } catch (e) {
+        // Silent catch.
```
yorkie: is the try-catch block really necessary? which case the `RegExp` will throw an error?

3y3: @yorkie , I'm not sure that this is a best time to review internal structure of this pr. At current step we need to decide how to avoid caching. Read my last comments.

Relative to your question:
```
new RegExp('abc[')
```

3y3: Anyway it would be nice if you'll help me with review!

a7madgamal:
@marcominetti I installed your fork and its working greatly but when i search and click on a line it always highlight the NEXT line!
I want you to confirm this and if it turns out to be a bug I think I can help you fix it quickly 
HINT:check the function `getLine(doc, n)` `return chunk.lines[n];` in the "codemerror.js" file should be n-1 if not n !==0

@3y3 I think it's ok for the search process to be a little bit slower and works like 10 files at a time then next 10 etc. for a basic fix. After all it's better than nothing!

3y3:
@a7madgamal , I agree that we must sacrifice performance for stability

marcominetti:
@a7madgamal yep the bug of next line is fixed within an unpublished version... I'll publish it after some cleanup...

marcominetti:
@3y3, I'm busy but I'll try to comment code just to share the pr/strategy... and I think the best way to improve stability and avoid memory leak is to directly check the filesystem if the file exists and, if not, trying to get through debuggerclient with requests "pagination"... do you agree?

marcominetti:
@3y3 hi, i'll have to work on this next weekend... just to confirm, i'll do:
- remove caching and
  - read script contents from fs
  - eventually pool by X when fs is not available/exist
- include fix for wrong-next-line

marcominetti:
@3y3 I've rebased against 0.11.1, reverted most of the code and switched to original functions extracted from frontend... It seems to work now. Performance are not exciting but no memory leaks. We can drop my custom implementation with regexp (commented code in SearchOverrides.js and _xxx "private" named functions in search.js). After review I'll fix the jshint stuff...

lib/search.js
```diff
@@ -0,0 +1,172 @@
...
+  for (var i = 0; i < query.length; ++i) {
+    var c = query.charAt(i);
+    if (regexSpecialCharacters.indexOf(c) != -1)
+      regex += "\\";
+    regex += c;
+  }
+  return new RegExp(regex, flags || "");
+}
+
+exports.performSearchInContent = function(content, query, caseSensitive, isRegex)
```
marcominetti: @3y3 this code is adapted from /front-end/common/ContentProvider.js implementation...

lib/search.js
```diff
@@ -0,0 +1,172 @@
...
+{
+  var regex = createSearchRegex(query, caseSensitive, isRegex);
+
+  var isMinified = false;
```
marcominetti: @3y3 I've experienced problem at front-end with long minified lines rendering... the client freezes... so I try to guess if the file has been minified in one line... (does not solve all cases but most of them)

lib/search.js
```diff
@@ -0,0 +1,172 @@
...
+  }
+  
+  var result = [];
+  var lineEndings = getlineEndings(content);
+  var lineCount = lineEndings.length;
+  for (var i = 0; i < lineCount; ++i) {
+    var lineContent = getlineAt(content, i, lineEndings);
+    regex.lastIndex = 0;
+    if (regex.exec(lineContent)) {
+      if (isMinified === true && lineContent.length > 1024) {
```
marcominetti: @3y3 we could just check here for lineContent length and replace it with "... (too long line)"

lib/search.js
```diff
@@ -0,0 +1,172 @@
...
+  }
+  return result;
+}
+
+var SearchMatch = function(lineNumber, lineContent) {
+  this.lineNumber = lineNumber;
+  this.lineContent = lineContent;
+}
+
+var _createSearchRegex = function(query, caseSensitive, isRegex)
```
marcominetti: @3y3 functions below are my attempt to gain performance but I think the incresed complexity of regexps and the allocation of match results to get line numbers are crazy... so these lines could be deleted (are now here just for reference)

3y3:
Hello @marcominetti , sorry for ignoring your previous comment - I'm very busy at this time. I'll review your changes as soon as possible. I'm not sure that this will happen today, but definitely on this week.

grsabreu:
Waiting for this =) @3y3

marcominetti:
Hi @grsabreu, in the weekend I'll check the status against latest commits to the master branch... are you experiencing specific problems?

3y3:
@marcominetti , please don't miss time at current weekend. I'll release "Node Inspector Next" at this month (I hope) with new frontend and completely reimplemented server. So it's not reasonable to rebase at current master.

Node Inspector Next breaks compatibility with oldest node versions (compatibility starts with 4.*), but implements lot of new features.

I'm sorry that I pending this pr, but each time then I start to review it, there happens new broken node release.

marcominetti:
@3y3 cool! Do you need help on nin?

a7madgamal:
any updates?

PR #622, opened 2015-04-22
Fix console in debug-brk mode

3y3:
On node 0.12 we can't use any custom commands (like 'Console.enable') if we stopped on first line.
This is a bug that can be fixed only in node source code.

Here I delete complicated console wrapping - we don't need to send 'Console.enable' to wrap console now.

This doesn't fixes initial problem. For example on 0.12 we can't get snapshot if we paused on first line. We need to send `continue` before using any custom command.

@bajtos , please review.

### DIFF
--- a/lib/ConsoleAgent.js
+++ b/lib/ConsoleAgent.js
@@ -11,7 +11,6 @@ var convert = require('./convert');
 function ConsoleAgent(config, session) {
   this._noInject = config.inject === false;
   this._injected = false;
-  this._consoleEnabled = false;
   this._debuggerClient = session.debuggerClient;
   this._frontendClient = session.frontendClient;
   this._injectorClient = session.injectorClient;
@@ -38,8 +37,6 @@ ConsoleAgent.prototype._inject = function(injected) {
       this._injected = !error;
 
       if (error) return this._frontendClient.sendLogToConsole('error', error.message || error);
-
-      if (this._consoleEnabled) this._debuggerClient.request('Console.enable');
     }.bind(this)
   );
 };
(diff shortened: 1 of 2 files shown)


bajtos:
I don't fully understand the consequences of this change, but I don't see any obvious problem either. LGTM.

As for the problem "we can't use any custom commands when stopped on the first line" - is it because your injected code is running in the same V8 isolate as the debugged application and thus when the app is paused, the injection cannot run?

It should be relatively easy to fix this problem in v8-debug, you need to run the server (listener) in a new isolate. I don't understand all details yet, see [src/debug-agent.cc](https://github.com/iojs/io.js/blob/e61ee49c7a2f577eb1338ee35f8768d79010ebc9/src/debug-agent.cc) and [lib/_debug_agent.js](https://github.com/iojs/io.js/blob/e61ee49c7a2f577eb1338ee35f8768d79010ebc9/lib/_debug_agent.js) to get started.

3y3:
>As for the problem "we can't use any custom commands when stopped on the first line" - is it because your injected code is running in the same V8 isolate as the debugged application and thus when the app is paused, the injection cannot run?

I don't think so. At the current time I work on next theory:
V8 (used in node 0.12) returns new instance of debug context each time when it's possible. This behavior was fixed in `v8-debug` [here](https://github.com/node-inspector/v8-debug/blob/master/v8-debug.js#L103-L123). So, how you can see I define debug event listener, that changes debug context behavior to be a persistent value.
Unfortunately at the time when I setup debug event listener, one instance of debug context was already defined. I don't have access to this instance, so can't extend it by custom commands.
After `continue+pause` we'll receive expected (persistent modified) debug context.

To fix this I need to move [this lines](https://github.com/node-inspector/v8-debug/blob/master/v8-debug.js#L120-L122) to node source code.

PR #627, opened 2015-04-29
Plugins system: extending front-end

3y3:
1. `Front-end` prefixed commits - example of new entry point for plugins.
2. `Plugins` prefixed commits - implement first step of plugin system.
On this step was implemented system for extending front-end.
Run from `node-inspector` directory:
  ```bash
  git clone https://github.com/3y3/ni-ansi-support.git plugins/ansi-support

  node bin/node-debug --plugins --no-b test
  ```
3. Was deprecated script generation on updating front-end. All scripts generated on startup.
4. New option `--plugins` with default `false`. When all parts of plugin system will be implemented, default will be changed to `true`

In next prs I'll implement also:
- `SessionModules` class and `SessionAgents` class for extending backend
- `ni install` script
- Maybe. Security policy configuratin

Current implementation allow to completely modify front-end


3y3:
Ready to review. @bajtos .

@marcominetti , your help in review is also interesting.

marcominetti:
Great @3y3, I'll do it later in the evening.

bajtos:
@3y3 could you please add a new section to README.md explaining how to use plugins? It would be great to have some documentation for plugin authors too, I think it's better to move the docs to a different place (either a new md file or perhaps a wiki page).

I don't have enough time to review this PR in detail. If it works for you and for @marcominetti, then I am fine with merging it as it is.

3y3:
@bajtos , there is one wrong place:
`InspectorBackendCommands.js` and `SupportedCSSProperties.js` moved to `front-end-node` folder (you requested this movement some prs ago). But it doesn't work as expected.
Will be it acceptable to add new express `get` statement like this:
```js
app.get('/InspectorBackendCommands.js', path.join(OVERRIDES, './InspectorBackendCommands.js'));
app.get('/SupportedCSSProperties.js', path.join(OVERRIDES, './SupportedCSSProperties.js'));
```
Or you see more elegant way?

3y3:
> could you please add a new section to README.md

I prefer to add new document in `docs` folder with main info about plugins.
+ Wiki pages
+ Link from readme.md

auchenberg:
Question: Is the idea to Fork the DevTools front-end in order to support
node-inspector specific plug ins?

( I'm on my mobile now so I can't review the code, so sorry if this is
clear when seeing the code)

DevTools already support extensions via the chrome extension APIs, so this
makes me wonder...

/k

bajtos:
@3y3 
> InspectorBackendCommands.js and SupportedCSSProperties.js moved to front-end-node folder (you requested this movement some prs ago). But it doesn't work as expected.

Could you be more specific please? What does it mean it does not work as expected? Are there other copies of these two files that are served instead of the desired ones? What is the underlying reason for that?

> I prefer to add new document in docs folder with main info about plugins.

Sounds good :+1:

3y3:
@auchenberg ,
> Question: Is the idea to Fork the DevTools front-end in order to support
node-inspector specific plug ins?

No, we won't to design our own plugins system for DevTools.

This is a first step of complex plugin system for Node Inspector.
We have three levels, where we want to inject code: frontend, inspector's backend, application.
This PR implements entry point to be able to load plugins to frontend.
So you have full access to DevTools sources, i.e. you can use `chrome extension APIs` or modify internal devtools methods.

3y3:
@bajtos ,

We use `front-end-node` as a folder with all overrides, but we store all overrides relative to `/node` path.

Main reason to remove `InspectorBackendCommands` from `front-end` - cleanup this folder from any changes. After this step there will no files, which we need to copy to `front-end` folder using `update-front-end` script.

But if we simple move `InspectorBackendCommands` to `front-end-node` folder, this doen't work, because it will be resolved as `/node/InspectorBackendCommands.js`, but we expect `/InspectorBackendCommands.js`

bajtos:
> We use front-end-node as a folder with all overrides, but we store all overrides relative to /node path.

Ah, that makes sense now.

The reason why I dislike manual enumeration of files via `app.get` is that one has to change the code whenever a new file is added. That's why I prefer to serve a full folder instead.

It make sense to keep front-end static without any NodeInspector-generated code. In that case, I am proposing to create a new folder for the generated files and serve it at the root URL.

```js
app.use('/node', express.static(OVERRIDES));
app.use(express.static(GENERATED));
app.use(express.static(WEBROOT));
```

Let's not dwell on this detail too much. I am ok with whatever solution you prefer.

auchenberg:
@3y3 Got it. I'll keep an eye on this progress then.

3y3:
I fixed last comments and merged it as is without readme.
At current time this feature is disabled by default.
I merged it because there is some changes which I want to see in master (all related with dynamic `front-end` generation)
I need some other updates (`update frontend` and `inject more code to app`) before I'll can completely finish plugin system.
After main part of plugin system will be finishe and before releasing I'll write readme.

PR #634, opened 2015-05-05
Use `wss:` WebSocket protocol if front-end is using SSL

### LINKED ISSUE #633: SSL broken in v0.10
After updating to Node Inspector v0.10, I was no longer able to debug over HTTPS. The browser console showed the following error:

```
Mixed Content: The page at 'https://172.17.42.1:8080/debug?ws=172.17.42.1:8080&port=5858' 
was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 
'ws://172.17.42.1:8080/?port=5858'. 
This request has been blocked; this endpoint must be available over WSS.
```

This broke after cbafc35, where WebSocket connections were changed to be handled by Blink's dev tools implementation.

But the problem is, Blink's implementation lacks support for the `wss` protocol... As seen [here](https://github.com/node-inspector/node-inspector/blob/master/front-end/main/Main.js#L267).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13217269-ssl-broken-in-v0-10?utm_campaign=plugin&utm_content=tracker%2F195817&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F195817&utm_medium=issues&utm_source=github).
</bountysource-plugin>

adambiggs:
This fixes #633 by hacking the upstream Blink source... I tried to add this in `MainOverrides.js` at first, but that didn't work because `WebInspector.Main.prototype._createConnection()` is called before `MainOverrides.js` is sourced...

I'm not really happy with this solution, but I've spend too much time trying to fix this as it is :(

----
Update: Fixed properly with help from @3y3 :beers:

### DIFF
--- a/front-end-node/NodeInspectorOverrides.js
+++ b/front-end-node/NodeInspectorOverrides.js
@@ -1,5 +1,5 @@
 /*jshint browser:true, nonew:false*/
-/*global WebInspector, Runtime, InspectorFrontendHost*/
+/*global WebInspector, Runtime, InspectorFrontendHost, InspectorBackendClass*/
 WebInspector.NodeInspectorOverrides = function() {
   this._overridenStrings = {
     'Developer Tools - %s': 'Node Inspector - %s',
@@ -7,9 +7,10 @@ WebInspector.NodeInspectorOverrides = function() {
   };
   this._overrideMainScriptType();
   this._overrideUIStrings();
+  this._overrideWebSocketCreate();
 
   this._setWorkerTitle();
-  
+
   this._mergeConnectionQueryParams();
 
   this._openMainScriptOnStartup();
@@ -42,6 +43,20 @@ WebInspector.NodeInspectorOverrides.prototype = {
     };
   },
 
+  _overrideWebSocketCreate: function() {
+    InspectorBackendClass.WebSocketConnection.origCreate =
+      InspectorBackendClass.WebSocketConnection.Create;
+
+    InspectorBackendClass.WebSocketConnection.Create = function(url, onConnectionReady) {
+      var args = Array.prototype.slice.call(arguments);
+      // If front-end is loaded via HTTPS, WebSocket protocol needs to be
+      // changed to WSS. Otherwise the browser will block the connection.
+      var protocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
+      args[0] = url.replace(/^ws:/, protocol);
+      this.origCreate.apply(this, args);
+    };
+  },
+
   _setWorkerTitle: function() {
     // Front-end uses `eval location.href` to get url of inspected page
     // This does not work in node.js from obvious reasons, and cause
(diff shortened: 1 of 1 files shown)


3y3:
@adambiggs , thanks for bug report and pr.
Now my turn to lose some time trying to override this using `front-end-node` folder =)
Anyway solution will be very hackly...
I'll return to review from some time.

3y3:
So [this](https://github.com/node-inspector/node-inspector/blob/master/front-end/sdk/InspectorBackend.js#L677-L680) looks like a nice place to override. Are you want to research this or you leave it for me?

adambiggs:
@3y3 oh nice! I'll update the PR.

front-end/sdk/InspectorBackend.js
```diff
@@ -676,6 +676,8 @@ InspectorBackendClass.WebSocketConnection = function(url, onConnectionReady)
  */
 InspectorBackendClass.WebSocketConnection.Create = function(url, onConnectionReady)
 {
+    var protocol = window.location.protocol === "https:" ? "wss:" : "ws:";
+    url = protocol + url.substr(3);
```
3y3: I don't like magic number `3`.
Keep in mind that we update `front-end` folder automatically, the future webkit realisation can contain on this place `wss`, then result can be:
```js
// url == 'wss://127.0.0.1' => true
url = protocol + url.substr(3); // => wsss://127.0.0.1
```

Next code will be more safely:
```js
url = url.replace(/^ws:/, protocol);
```

3y3:
Reviewed. 
The last step is to move this modification to [NodeInspectorOverrides.js](https://github.com/node-inspector/node-inspector/blob/master/front-end-node/NodeInspectorOverrides.js) (something like `_checkWebSocketProtocol`), which depends on `sdk` only. So we avoid:
>WebInspector.Main.prototype._createConnection() is called before MainOverrides.js is sourced...


Read also my inline comment.

front-end/sdk/InspectorBackend.js
adambiggs: Good point, fix incoming (force-pushed).

adambiggs:
I moved everything into `NodeInspectorOverrides.js`, and tried to follow the existing conventions in that file. Does that look like a good solution?

3y3:
Looks great! Please fix some linter errors and I'll merge this.
To see linter's errors you can run `npm test` or look at [travis log](https://travis-ci.org/node-inspector/node-inspector/jobs/61496239)

adambiggs:
@3y3 done!

3y3:
@adambiggs , great thanks for contribution. :beers:

mamacdon:
Brilliant, thanks for the fix!

PR #637, opened 2015-05-07
Fix breakpoints in required files

3y3:
@bajtos , looking on node 0.12 I found real reason why we can't previously pause on first line.
This is because we don't receive `afterCompile` event, that registers scripts in ScriptManager.
Isntead of this we receive `compileError` event every time.

So in this pr I deprecate event caching feature in BreakEventHandler. 
I implemented `_onBreak` as an async function, this provide more powerful control flow and allow me to asynchronously receive script source (if it needed) directly from `_onBreak`


lib/DebuggerAgent.js
```diff
@@ -105,7 +105,7 @@ DebuggerAgent.prototype = {
     this._debuggerClient.request(
       'scripts',
       {
-        includeSource: true,
+        includeSource: false,
```
3y3: I don't know why we need to receive sources here.
I can revert this line or move it to isolated commit.

3y3:
I targetted issues that will be (or can be) solved by this pr.

3y3:
Ping @bajtos . Is it possible to review it on this week and release 0.10.1?

lib/DebuggerAgent.js
bajtos: See https://github.com/node-inspector/node-inspector/commit/39fe73b620738381efbb9db3239e6eba074de1cc and #307

If the new front-end does not need `_getSourceMapUrl` anymore, then it's safe to change this flag back to `false`. Otherwise it would be better to investigate this more.

Regardless of the solution, please consider making the change in a new pull request and explaining the context of the change. It will make our life easier later in the future, when we will look at this line again and scratch our head trying to understand why it was changed.

bajtos:
> This is because we don't receive afterCompile event, that registers scripts in ScriptManager.
> Isntead of this we receive compileError event every time.

I find this rather weird. Are you sure it's not a bug in Node and/or V8, or perhaps in the way how we are using the debugger protocol? Can we reproduce the problem in such way that allows us to report this as a bug to Node.js and/or io.js projects?

> in this pr I deprecate event caching feature in BreakEventHandler. 
> I implemented _onBreak as an async function, this provide more powerful control flow and allow me to asynchronously receive script source (if it needed) directly from _onBreak

Cool, this sounds like a better solution than the event caching mechanism! I am afraid I don't have enough time to fully understand the scope of these changes. If it works for you then I am ok with it.

One scenario to consider and/or test manually:
 - Start Node Inspector with `--no-preload`
 - Start the debugged process via `node --debug`, make sure the debugged process loads (requires) some other scripts
 - Start the debugging session in browser. Verify that scripts loaded before the debugger was attached are correctly displayed in the UI.

I thing [sourcemaps sample app](https://github.com/node-inspector/node-debug-samples/tree/bef5febd83d1e401abeb93933c5e033d14dedc16/sourcemap) may be a good app to debug.

bajtos:
FWIW, the CI build is failing for io.js@2.0. I think the upgrade of V8 changed the data reported by the debugger.

3y3:
> I find this rather weird. Are you sure it's not a bug in Node and/or V8, or perhaps in the way how we are using the debugger protocol? Can we reproduce the problem in such way that allows us to report this as a bug to Node.js and/or io.js projects?

Already reported https://github.com/joyent/node/issues/25266.

I have [confirmation in gitter chat](https://gitter.im/node-inspector/node-inspector?at=554ca56c5e423e6573fce2e8), that this pr solves original problem (tested without `--no-preload`)
I'll follow your test list and if all works on `--no-preload` flag, is it ready to merge?

>FWIW, the CI build is failing for io.js@2.0. I think the upgrade of V8 changed the data reported by the debugger.

There is a lot of problems. I tried to fix tests for iojs some times, but this requires a deeper digging.

bajtos:
> I have confirmation in gitter chat, that this pr solves original problem (tested without --no-preload)
I'll follow your test list and if all works on --no-preload flag, is it ready to merge?

Sure, no problem. I think it would be best to test both `--no-preload` and source maps.

bajtos:
@3y3 Thank you for taking care of this and keeping the project going!

3y3:
Tested on `sourcemap`. Works fine.
Reverted `includeSource` option in `DebuggerAgent`.

PR #647, opened 2015-05-18
Fix source maps

3y3:


lib/ScriptFileStorage.js
```diff
@@ -58,19 +58,24 @@ $class.save = function(path, content, callback) {
...
-      // remove shebang
-      content = content.replace(/^\#\!.*/, '');
-
-      var source = MODULE_HEADER + content + MODULE_TRAILER;
-      return callback(null, source);
-    }
-  );
+  var scriptId = this._scriptManager.findScriptIdByPath(path);
+  if (scriptId != null) {
+    this._scriptManager.getScriptSourceById(scriptId, callback);
```
marcominetti: Pay attention that result from this._scriptManager.getScriptSourceById is an object not the content...

```javascript
this._scriptManager.getScriptSourceByIdFromCache(scriptId, function(error,data) {
        if (err) return callback(err);
        callback(null, data.scriptSource);
      });
```

lib/ScriptManager.js
```diff
@@ -161,7 +193,8 @@ ScriptManager.prototype = Object.create(events.EventEmitter.prototype, {
             }
             debug('sourceMapUrl for script %s:%s is %s',
               v8data.id, localPath, sourceMapUrl);
-            inspectorScriptData.sourceMapURL = sourceMapUrl;
+            inspectorScriptData.sourceMapURL = this._checkInlineSourceMap(sourceMapUrl);
+            console.log(inspectorScriptData.sourceMapURL);
```
bajtos: Forgotten console.log?

lib/ScriptManager.js
```diff
@@ -4,7 +4,8 @@ var events = require('events'),
     convert = require('./convert.js');
 
 // see Blink inspector > ContentSearchUtils.cpp > findMagicComment()
-var SOURCE_MAP_URL_REGEX = /\/\/[@#][ \t]sourceMappingURL=[ \t]*([^\s'"]*)[ \t]*$/m;
+var SOURCE_MAP_URL_REGEX = /\/\/[@#][ \t]sourceMappingURL=[ \t]*([^\s'"]*)[ \t]*$/m,
+    INLINE_SOURCE_MAP_REGEX = /^(data:application\/json;base64,)(.*)$/;
```
bajtos: I think your regexp is too strict. In the past, I wrote data-URI parser specifically for Node Inspector, see https://www.npmjs.com/package/strong-data-uri. Please consider using that module instead of a re-implementing the parser again.

bajtos: I think your regexp is too strict, the URI can have other header and still stay valid:

```
data:application/json;charset=utf-8;base64,...data...
data:application/json,...url-encoded-data...
```

In the past, I wrote a data-URI parser specifically for Node Inspector, see https://www.npmjs.com/package/strong-data-uri. Please consider using that module instead of re-implementing the parser again.

3y3:
@bajtos , work in progress. This is not ready to review.
Big part of this pr will be deprecated.

Thanks for targetting to `strong-data-uri`

marcominetti:
@3y3 @bajtos is reasonable to include browserified version of strong-data-uri or any replacement at front-end too for the https://github.com/marcominetti/node-inspector/blob/fix-sourcemaps-alternative/front-end-node/main/MainOverrides.js#L48?

bajtos:
> is reasonable to include browserified version of strong-data-uri or any replacement at front-end too for the https://github.com/marcominetti/node-inspector/blob/fix-sourcemaps-alternative/front-end-node/main/MainOverrides.js#L48?

IMO, it would be best to share the code handling sourcemaps between frontend and backend, so that they don't get out of sync. Using browserify is a possible solution.

Another option is to store inline sourcemaps on the server (backend) and replace the data URL in the source code with a URL pointing to a custom server handler that can serve the source map. That may make more sense in the case when the backend is already performing some sort of transformation/preprocessing of inline source maps.

Issue #628, opened 2015-04-30
Can't set breakpoint via web inspector UI (iojs with babel)


th0r:
I've recorded short video, displaying this bug: http://youtu.be/odIFQ1DNCAc
I'm using `iojs v1.8.1` with `babel 5.1.13`, which transpiles scripts via `require` hook and generates inline sourcemaps.
If you need some additional info, just ask.

cr0cK:
I have exactly the same problem and it's annoying.

I start the serveur with :
./node_modules/.bin/babel-node --debug --stage 1 index.js

rollokb:
I also have this problem. I need a fix urgently

3y3:
Looks like a duplication of https://github.com/node-inspector/node-inspector/issues/580

3y3:
Provide me please very small example to reproduce this problem.

cr0cK:
I push a small example where I can reproduce the bug: https://github.com/cr0cK/node-inspector-issue628

Tested version are in the README of the example.

marcominetti:
Thanks @cr0cK... I'll investigate during the we...

marcominetti:
Ok, problem found... It's related with file naming... Seems node-inspector is not able to handle the case when the compiled and loaded script uri (e.g. /home/marco.minetti/Desktop/node-inspector-issue628/app/classes.js) is identical to the source file uri embedded in the sourcemap...

I'll try to fix it during the weekend... just as workaround, in my projects I force babel and acorn to use fake (virtual) source paths to keep compiled and source uri different... but with inline sourcemap this is not the case...

marcominetti:
@3y3 Yuri, when you have time can you send me the repo where the blink tools source code is available... maybe they have fixed something after new frontend merge...

3y3:
@marcominetti , I already have in plans to test front-end update (with new front-end system, this promises to be simple), so, if you want it soon, I'll make it on this evening or on weekend.

marcominetti:
@3y3 could be useful, but I've created a workaround with duplication of entries in resource tree, I mean you get two files (file://classes.js - the original - and built://classes.js - the transpiled that could be also hidden)... let me know if it sounds reasonable for you in here or wants to delve into it...

3y3:
@marcominetti , can you open a pr to look on progress?

marcominetti:
yep, i'll do asap during the weekend... ;)

3y3:
Fixed in 0.10.2.
[...]

PR #648, opened 2015-05-18
Fix sourcemaps re-fetched from backend

marcominetti:
@3y3 This keeps proper source file names, but creates fake "loaded" scripts... I think the problem should be addressed at transpiling level (Babel, etc): I don't think it's fair to generate in-memory sources with same file names... (maybe in sourcemap v4 spec could be added a proper way to handle in-memory transpiling)...

### DIFF
--- a/front-end-node/main/MainOverrides.js
+++ b/front-end-node/main/MainOverrides.js
@@ -4,6 +4,7 @@ WebInspector.MainOverrides = function() {
   this._unregisterShortcuts();
   this._allowToSaveModifiedFiles();
   this._reloadOnDetach();
+  this._avoidSourceMapFetchWhenInline();
 };
 
 WebInspector.MainOverrides.prototype = {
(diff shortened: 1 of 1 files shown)


3y3:
Ok. At current step I like my changes in `ScriptFileStorage` and your changes in `front-end`.
Next step - discussion about renaming files...
I think that all renamings is out of scope of node-inspector.
We can't solve existed situation correctly (this is an issue for babel), but we also can't to add dirty fix.

If we leave only changes in `ScriptFileStorage` and in `front-end` this would work for correct sourcemaps. @marcominetti , are you agree?

```
root
// loading script source

root
-> source.js
// processing sourceMapUrl

root
-> source.js + source.map
// processing source paths is source map

root
-> source.js + source.map
-> source.js (from source map)
// Unresolved situation
```


One other thing. There are inlined sources. So at current time we prevent requiring of inline sourcemap, but we can also prevent requiring of inline sources which stored in inline sourcemap (or it already works?).

[Source Map Revision 3 Proposal](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit#heading=h.hval2vfopveu)

marcominetti:
`If we leave only changes in ScriptFileStorage and in front-end this would work for correct sourcemaps. @marcominetti , are you agree?`

I do agree. ;)

`One other thing. There are inlined sources. So at current time we prevent requiring of inline sourcemap, but we can also prevent requiring of inline sources which stored in inline sourcemap (or it already works?).`

I've not tested specifically but it should already work... I use babel and uglify to build an app boot pack for node and front-end resource tree correctly lists original (virtual) source code... with not-inlined sourcemap the code is read from sourcemap sourcesContent property...

marcominetti:
@3y3 please think about changing the "map" prefix to extension for inline mapped files... having a real source file with name "xxx.map.js" is not impossible... maybe a __map__?

marcominetti:
@3y3 I reverted my changes to backend and added a fix to latest front-end that was breaking not-inlined sourcemap.

marcominetti:
Fixes #628 and #580 ... original source code for "name.ext" will be available as "name.source.ext"

3y3:
@marcominetti , read also https://github.com/babel/babel/pull/1587

marcominetti:
@3y3 that's exactly the workaround I use within my apps to avoid the issue.. I'll reduce this pr to front-end code only, ok?

marcominetti:
@3y3 kept only the front-end fix, I've not yet identified if there's any data-uri parser that supports other then application/json;base64... IMO, we could merge and than extend support to full data-uri spec in the future...

3y3:
@marcominetti , sorry for pending this. I don't have a time to work on source maps on this week, but I follow the work.
We also discussed about this bug with Bajtos in gitter chat, so I need to fix and extend my part of commits.

Don't worry about this now. I'll rebase and merge all important parts and I'll present it for your review.

3y3:
@marcominetti , I landed it as be64d801ea8bbf418ba316a23610b16ad1a4f1cb with some minor changes.

Thank you for contribution!

Issue #341, opened 2014-04-05
Feature Request: Network Tab?


go1dfish:
Not sure how realistic this is as I'm not familiar with the internal structure of node.js wrt http requests.

But if it were possible for node-inspector to implement something like Chrome's Network tab for the debugging of outgoing/incoming http requests this would be very powerful.

hshepherd:
+1

petercsoka:
+1

jkrems:
Some older implementation open for grabs/porting:

* The injected code: https://github.com/jkrems/bugger/blob/master/src/bug-script/probes/network.coffee
* The inspector/agent side: https://github.com/jkrems/bugger/blob/master/src/domains/Network.coffee

marcominetti:
:+1:

jkrems:
(Sorry, didn't see that the file was actually already linked above.)

jkrems:
The newer / nicer version is here: https://github.com/buggerjs/bugger-agents/blob/master/lib/agents/network.probe.js

Working on porting it now. But just in case I get distracted and someone else wants to take this on.

marcominetti:
:+1:

atif089:
@jkrems I was building a web scraper and I have a read world scenario to try this out. Can you please help me how to merge your code with my copy of module to try this? (I'm assuming I can't do a git cherry-pick or something like that because the module itself is not a git repo)


Alternatively, the repo where I want to try this is here https://github.com/atif089/RandomActsOfTech/tree/master/InstaUserScraper

3y3:
Closed in #677

PR #655, opened 2015-05-27
Basic network request inspection

jkrems:
Original issue: https://github.com/node-inspector/node-inspector/issues/341

Based on [bugger's network probe](https://github.com/buggerjs/bugger-agents/blob/master/lib/agents/network.probe.js). The rest is basically one big copy&paste job of `ConsoleAgent`.

I kept this PR as small as possible because messing with the `http` module is prone to break people's code in edge cases. Tests against real-life code more than welcome.

### DIFF
--- a/lib/networkInjection.js
+++ b/lib/networkInjection.js
@@ -1,0 +1,137 @@
+// This function will be injected into the target process.
+module.exports = function(require, debug, options) {
+  var url = require('url');
+  var STATUS_CODES = require('http').STATUS_CODES || {};
+
+  var lastRequestId = 0;
+
+  function _makeStackTrace(fn) {
+    var backup = Error.prepareStackTrace;
+    Error.prepareStackTrace = function(_, stack) { return stack; };
+    var err = new Error();
+    Error.captureStackTrace(err, fn);
+    var stack = err.stack;
+    Error.prepareStackTrace = backup;
+
+    // TODO: handle empty stack (breaks frontend)
+    return stack.map(function(frame) {
+      return {
+        url: (frame.getFileName() && 'file://' + frame.getFileName()) || frame.getEvalOrigin(),
+        functionName: frame.getFunctionName(),
+        lineNumber: frame.getLineNumber(),
+        columnNumber: frame.getColumnNumber()
+      };
+    });
+  }
+
+  // Dirty trick reaching deep into core to access the headers about to be sent
+  // Likely to break eventually.
+  function _renderHeaders(req) {
+    var headers = req._headers;
+    if (!headers) return {};
+
+    var out = {}, names = Object.keys(headers);
+    names.forEach(function(name) {
+      out[req._headerNames[name]] = headers[name];
+    });
+
+    return out;
+  }
+
+  function _removeArraysFromHeaders(headers) {
+    var out = {}, names = Object.keys(headers);
+    names.forEach(function(name) {
+      var value = headers[name];
+      if (Array.isArray(value)) {
+        value = value.join('\n');
+      }
+      out[name] = value;
+    });
+    return out;
+  }
+
+  var DEFAULT_MIME = 'text/plain';
+  function _getMimeType(headers) {
+    if (typeof headers['content-type'] === 'string') {
+      return headers['content-type'].split(';')[0] || DEFAULT_MIME;
+    }
+    return DEFAULT_MIME;
+  }
+
+  function _patchLib(protocolLib) {
+    var originalRequest;
+    originalRequest = protocolLib.request;
+    protocolLib.request = request;
+
+    function request(options, onResponse) {
+      if (typeof options === 'string') {
+        options = url.parse(options);
+      }
+
+      var initiatorStackTrace = _makeStackTrace(request);
+
+      var requestId = '' + (++lastRequestId);
+      var loaderId = requestId;
+      var documentURL = url.format(options.uri || options);
+      var clientRequest = originalRequest(options);
+      var originalEnd = clientRequest.end;
+      clientRequest.end = function end() {
+        var data = {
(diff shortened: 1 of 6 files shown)


lib/networkInjection.js
```diff
@@ -0,0 +1,137 @@
+// This function will be injected into the target process.
+module.exports = function(require, debug, options) {
+  var url = require('url');
+  var STATUS_CODES = require('http').STATUS_CODES || {};
+
+  var lastRequestId = 0;
+
+  function _makeStackTrace(fn) {
```
jkrems: `ConsoleAgent` uses a magical offset, might be worth to use the `fn` argument there as well instead.

3y3:
@jkrems , great job!

After fast review I have one idea:
```js
var request = require('http').request;
```
If this code was called before wrapping, we can't handle this requests, because user already has closure on original `request` func.
It may be reasonable to wrap something more private. Maybe [http.ClientRequest](https://github.com/joyent/node/blob/master/lib/http.js#L48)

jkrems:
Well, `request` has a closure over `ClientRequest`. So if a user already has a closure over `request`, they also have one over `ClientRequest` (at least indirectly). It might be possible to change the code of `request` itself at runtime, but that would make me a bit nervous. ;)

3y3:
> Well, request has a closure over ClientRequest.

Ok, you are right here, but I don't stop to search an other place =)

So, I found [_deferToConnect](https://github.com/joyent/node/blob/master/lib/_http_client.js#L511), it exists on all versions. (We shouldn't worry about compatibility with future versions here, only check that this function exists. I know what `http` module can be completely rewritten in nearest time.)

There are all that we need for `data` construction (correct me if I'm wrong), but making stack trace will be more difficult. In this case we need to wrap only one place:
```js
function wrapHttpRequests() {
  var proto = require('http').ClientRequest.prototype;
  proto._deferToConnect = (function(fn) {
     var lastRequestId = 0;
     // ...
     return function handleHttpRequest() {
       debug.command('Network.requestWillBeSent', constructRequestData(this));
       this.once('response', handleHttpResponse);
       fn.apply(this, arguments);
     }
  } (proto._deferToConnect));
}
```

How about this?

jkrems:
Hehe, nice find. I didn't look too deeply into it but it doesn't look like `options` is stored anywhere on the instance of `ClientRequest`. So if we'd want to correctly reconstruct the original URL (and that's only one of the pieces of data) we'd need:

1. Find out the hostname. The client request doesn't store it. `Host` header is unreliable because it doesn't have to match the actual host (e.g. host-spoofing)
2. Find out the remote port. From reading the constructor I don't even know where to start to get that one, especially given that we'd only want to send `requestWillBeSent` if a dns lookup fails etc.. Failing requests are as or maybe even more interesting.

That last part also brings up the question - is there a guarantee that `_deferToConnect` is called? What if socket pooling is used and a hard timeout on the request to abort it if a socket can't be acquired after a few millis?

Sorry if I sound defensive here. I totally get that you'd love to get a more reliable/flexible http instrumentation working. It's just that the code in this PR (or at least variations of it) is something that I'm fairly confident works and is reliable to the point where I know of apps using it in production for outgoing http request logging. I could try to dig deeper into http core and research if there's an alternative that also handles the case where someone already has a reference to `request`. But unfortunately I won't get to around to that anytime soon. :( That being said - I'm happy to test and give feedback on other solutions!

GingerBear:
I am excited to see this finally coming. This is great! So I tested the against one of our project, very glad to see the network requests showing up in the network tab!

I also noticed some potential bugs
- network frontend stop working if header value is not a string (eg. content-length). network manager is calleing a split on header value [here](https://github.com/node-inspector/node-inspector/blob/master/front-end/sdk/NetworkManager.js#L162)
- when I restart node-inspector, the network request is doubled.
- the `options` in the hijacking request seems not quite matching the one in [http request api](https://nodejs.org/api/http.html#http_http_request_options_callback), which leads to hostname showing as the request url.

Hope this helps!

jkrems:
@GingerBear Thanks for giving it a spin! Will look into those issues!

jkrems:
Quick update: Pushed an update that adds the missing string cast for the headers. Also included a change to explicitly remove the instrumentation on disconnect, hoping that will fix the doubling issue. Unfortunately I wasn't able to reproduce it myself.

@GingerBear Do you have an example for the `options` object where the url is messed up? I'm assuming it's caused by the `host`/`hostname` weirdness in `url.format`. But I can't recall exactly how to trigger it.

GingerBear:
@jkrems sure, here is an example. May not be the best one, but you can see only www.google.com show in the network tab when you go to http://localhost:4000/intl/en/about/ . Btw, it seems to be fine using [request](https://github.com/request/request) to send request. 
I am still seeing the double logs. For the same example, refresh the inspector couple times and load the web page again, I was able to see the doubling in the network tab.
```javascript
var http = require('http');

var server = http.createServer(function(req, res) {
  var request = http.request({
      host: 'www.google.com',
      method: 'GET',
      path: req.url 
  }, function (response) {
      var body;
      response.on('data', function (data) {
          body += data;
      });
      response.on('end', function (data) {
          res.end(body);
      });
  });
  request.end();
});

server.listen(4000, function(){
    console.log("Server listening on: http://localhost:%s", 4000);
});
```

3y3:
> I am still seeing the double logs.

@jkrems , don't worry about this. This is out of scope of your pr.

Your `.on('close')` handling is reasonable.

jkrems:
@GingerBear Thanks! The code wasn't handling that `protocol` is optional. Pushed an update that should fix this.

Sorry for the slow reaction time. Busy week.

jkrems:
Closing in favor of #677

PR #675, opened 2015-06-11
Fix console

3y3:


lib/InjectorServer.js
```diff
@@ -15,7 +15,9 @@ function injectorServer(options) {
     // Get previously cached mirror if existed
     mirror = resolveCachedMirror(cache, mirror);
 
-    var serializer = makeMirrorSerializer(true);
+    var serializer = makeMirrorSerializer(true, {
+      maxStringLength: 10000
```
3y3: We store this constant in two places at current time. (this is bad solution and I can rewrite it on something more complex, if you want)

3y3:
@bajtos , ready to review.

lib/InjectorServer.js
bajtos: Perhaps we should use `-1` instead? AFAIK that way V8 does not truncate the response values at all.

bajtos:
@3y3 I am afraid I don't have enough time to review this in detail. I don't see any obvious problem, if the patch works for you then I am fine with it. :shipit:

3y3:
> Perhaps we should use -1 instead

Everywhere? Or only for console? I'm :+1:  for `-1`

bajtos:
> Everywhere? Or only for console? I'm :+1: for -1

I don't have a strong opinion. Shall we try to use it everywhere and see if it breaks anything?

PR #677, opened 2015-06-14
[WIP] Alternative Network tab implementation

3y3:
@jkrems , thanks to your experience in #655 I added full network tab implementation.
Here I solve some problems of your pr:
- We can wrap code in `--debug` and `--debug-brk` mode
- We wrap only one function
- We send finished requests to frontend
- Added different identifiers for `requestId`, `connectionId`, `loaderId`.
- We collect stack trace closest to func called by user
- POST data captured

What is implemented:
- Basic http request support
- Request timings
- Request data capturing and cleaning (we store data in node-inspector)
- Backported `String.prototype.toBase64` from latest DevTools. This solves problem with inspection of data with non ascii caracters

What doesn't solved in this pr:
- [TODO](https://github.com/node-inspector/node-inspector/blob/master/lib/FrontendCommandHandler.js#L48-L51) not fixed. So we can't use `XHR breakpoint` section in source panel.
- I don't know that happens if anyone will use not default encoding in `req.write`
- I don't test keep alive requests
- Not implemented `Network + Console` behavior.

### DIFF
--- a/lib/Injections/NetworkAgent.js
+++ b/lib/Injections/NetworkAgent.js
@@ -1,0 +1,396 @@
+// This function will be injected into the target process.
+module.exports = function(require, debug, options) {
+  var http = require('http'),
+    https = require('https'),
+    EventEmitter = require('events').EventEmitter;
+
+  var addRequest = http.Agent.prototype.addRequest,
+      onSocket = http.ClientRequest.prototype.onSocket;
+
+  var lastRequestId = 0,
+      lastConnectionId = 0;
+
+  function timestamp() {
+    return Date.now() / 1000000;
+  }
+
+  function Timing() {
+    this._hrtime = process.hrtime();
+
+    this.json = {
+      requestTime: Date.now() / 1000,
+      proxyStart: -1,
+      proxyEnd: -1,
+      dnsStart: -1,
+      dnsEnd: -1,
+      connectStart: -1,
+      connectEnd: -1,
+      sslStart: -1,
+      sslEnd: -1,
+      serviceWorkerFetchStart: -1,
+      serviceWorkerFetchReady: -1,
+      serviceWorkerFetchEnd: -1,
+      sendStart: -1,
+      sendEnd: -1,
+      receiveHeadersEnd: -1
+    };
+  }
+
+  [
+    'proxyStart', 'proxyEnd',
+    'dnsStart', 'dnsEnd',
+    'connectStart', 'connectEnd',
+    'sslStart', 'sslEnd',
+    'serviceWorkerFetchStart', 'serviceWorkerFetchReady', 'serviceWorkerFetchEnd',
+    'sendStart', 'sendEnd',
+    'receiveHeadersEnd'
+  ].forEach(function(method) {
+    Timing.prototype[method] = function() {
+      var diff = process.hrtime(this._hrtime);
+      this.json[method] = diff[0] * 1e3 + diff[1] / 1e6;
+      return this;
+    };
+  });
+
+  function getStackTrace() {
+    var backup = Error.prepareStackTrace;
+
+    Error.prepareStackTrace = function(_, stack) { return stack; };
+    // We want to trim stack trace to display user caller function.
+    // We wait what request will be sent by one of functions in `callers` list.
+    // If function exists in stack trace, his length will be great than 0
+    // If there is no known caller func we sent full stack trace
+    var callers = [handleHttpRequest, http.request, http.get, https.request, https.get],
+      error,
+      stack;
+
+    while (!stack && callers.length) {
+      error = new Error();
+      Error.captureStackTrace(error, callers.pop());
+      if (error.stack.length) stack = error.stack;
+    }
+
+    if (!stack) stack = new Error().stack;
+
+    Error.prepareStackTrace = backup;
+
+    return stack.reduce(function(stack, frame) {
+      var fileName = frame.getFileName();
+
+      fileName = debug.convert.v8NameToInspectorUrl(fileName);
+
+      var url = fileName || frame.getEvalOrigin();
+
+      stack.push({
+        url: url,
+        functionName: frame.getFunctionName(),
+        lineNumber: frame.getLineNumber(),
+        columnNumber: frame.getColumnNumber()
+      });
+
+      return stack;
+    }, []);
+  }
+
+  function renderHeaders(request) {
+    var headers = request._headers;
+    if (!headers) return {};
+
+    return Object.keys(headers).reduce(function renderHeader(result, key) {
+      var headerName = request._headerNames[key];
+      result[headerName] = '' + headers[key];
+      return result;
+    }, {});
+  }
+
+  function getCorrectHeaders(response) {
+    return Object.keys(response.headers).reduce(function correctHeader(result, key) {
+      var value = response.headers[key];
+
+      if (Array.isArray(value))
+        value = value.join('\n');
+
+      result[key] = '' + value;
+      return result;
+    }, {});
+  }
+
+  function getStatusText(response) {
+    return response.statusMessage || http.STATUS_CODES[response.statusCode] || '?';
+  }
+
+  function restoreHeadersText(response) {
+    var headers = response.rawHeaders;
+    // TODO: there is no `rawHeaders` in node 0.10.*
+    // We can wrap parser to find raw headers, but it's not so useful to see original string
+    // So, maybe won't fix
+    if (!headers) return;
+
+    var headersPrefix = [
+      'HTTP/' + response.httpVersion,
+      response.statusCode,
+      response.statusMessage
+    ].join(' ');
+
+    var joinedHeaders = [headersPrefix];
+    while (headers.length)
+      joinedHeaders.push(headers.splice(-2).join(': '));
+
+    return joinedHeaders.join('\r\n') + '\r\n\r\n';
+  }
+
+  function getMimeType(response) {
+    if (typeof response.headers['content-type'] != 'string')
+      return 'text/plain';
+
+    return response.headers['content-type'].split(';')[0];
+  }
+
+  function constructRequestInfo(request) {
+    var requestId = request.__inspector_ID__;
+    var timestamp = request.__inspector_timing__.json.requestTime;
+    var url = request.__inspector_url__;
+
+    return {
+      requestId: requestId,
+      loaderId: process.pid + '',
+      documentURL: url,
+      type: 'XHR',
+      request: {
+        headers: renderHeaders(request),
+        method: request.method,
+        postData: '',
+        url: url
+      },
+      timestamp: timestamp,
+      initiator: {
+        stackTrace: getStackTrace(),
+        type: 'script'
+      }
+    };
+  }
+
+  function constructResponseInfo(response) {
+    var request = response.req;
+
+    return {
+      requestId: request.__inspector_ID__,
+      loaderId: process.pid + '',
+      timestamp: timestamp(),
+      type: 'XHR',
+      response: {
+        url: request.__inspector_url__,
+        status: response.statusCode,
+        statusText: getStatusText(response),
+        headers: getCorrectHeaders(response),
+        mimeType: getMimeType(response),
+        connectionReused: request.shouldKeepAlive,
+        connectionId: request.connection.__inspector_ID__,
+        encodedDataLength: -1,
+        fromDiskCache: false,
+        fromServiceWorker: false,
+        timing: request.__inspector_timing__.json,
+        headersText: restoreHeadersText(response),
+        requestHeaders: renderHeaders(request),
+        requestHeadersText: request._header
+      }
+    };
+  }
+
+  function constructFailureInfo(request, err, canceled) {
+    // NOTE: If there is no other `error` listeners
+    // handling of `error` event changes program behavior
+    // We won't to destroy app on each failed request,
+    // but we need to notify user about unhandled `error` event.
+    var unhandled = err && EventEmitter.listenerCount(request, 'error') === 0;
+    var errorText = (unhandled ? '(unhandled) ' : '') + (err && err.code);
+    return {
+      requestId: request.__inspector_ID__,
+      timestamp: timestamp(),
+      type: 'XHR',
+      errorText: errorText,
+      canceled: canceled
+    };
+  }
+
+  function handleHttpRequest(options) {
+    var timing = new Timing();
+    var protocol = options.protocol || this.agent.protocol || 'http:',
+      host = options.host,
+      port = options.port,
+      path = this.path;
+
+    this.__inspector_timing__ = timing;
+    this.__inspector_ID__ = '' + lastRequestId++;
+    this.__inspector_url__ = protocol + '//' + host + ':' + port + path;
+
+    var requestInfo = constructRequestInfo(this);
+
+    this.once('socket', handleSocket.bind(this, requestInfo));
+    this.once('response', handleHttpResponse);
+    this.once('error', handleFailure.bind(this, requestInfo));
+    handleRequestData.call(this, requestInfo);
+    handleAbort.call(this, requestInfo);
+
+    timing.dnsStart();
+  }
+
+  function handleRequestData(requestInfo) {
+    var oldWrite = this.write;
+
+    this.write = function(chunk) {
+      requestInfo.request.postData += chunk || '';
+      return oldWrite.apply(this, arguments);
+    };
+  }
+
+  function sendRequestWillBeSent(requestInfo) {
+    if (requestInfo.handled) return;
+    requestInfo.handled = true;
+
+    debug.emitEvent('Network.requestWillBeSent', requestInfo);
+    debug.emitEvent('Network._requestWillBeSent', {
+      requestId: requestInfo.requestId
+    });
+  }
+
+  function handleSocket(requestInfo, socket) {
+    socket.__inspector_ID__ = socket.__inspector_ID__ || '' + lastConnectionId++;
+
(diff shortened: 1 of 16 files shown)


3y3:
@jkrems ,
@marcominetti ,
@bajtos ,

please review this. This is a full implementaton of Network panel.

3y3:
force updated to fix tests.

lib/Injections/NetworkAgent.js
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    function handleRequestEnd(request) {
+      var oldWrite = request.write,
+          oldEnd = request.end;
+
+      var requestInfo = constructRequestInfo(request);
+
+      request.write = function(chunk) {
+        oldWrite.apply(this, arguments);
```
jkrems: It's a good practice when instrumenting functions to always return this original function's return value. In this case:

> The return value indicates if you should continue writing right now. If the data had to be buffered internally, then it will return false. Otherwise, it will return true.

See: https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback

lib/Injections/NetworkAgent.js
```diff
@@ -0,0 +1,300 @@
...
+        requestInfo.request.postData += chunk || '';
```
jkrems: It might be a good idea to at least call `(chunk || '').toString('utf8')` to support buffers*. I'm not sure how to handle non-text though (e.g. when uploading an image in the body, not using multi-part).

(*) Or to fully copy concat-stream's implementation of collecting an array with all chunks and then to use `Buffer.concat` or `[].concat`.

lib/Injections/NetworkAgent.js
3y3: Good point. Will be fixed.

lib/Injections/NetworkAgent.js
3y3: > It might be a good idea to at least call (chunk || '').toString('utf8') to support buffers

But `postData` defined as empty string, so we have auto conversion to string in `+=` operator. Correct me if I'm wrong here. And `buffer.toString` use `utf8` bu default.

3y3: > I'm not sure how to handle non-text though

non-text is a real problem for current implementation...
I don't know what will be preferred: reimplement this part or release it as is and wait for users feedback

lib/Injections/NetworkAgent.js
```diff
@@ -0,0 +1,300 @@
...
+  function wrapHttpRequests() {
+    var lastRequestId = 0,
+        lastConnectionId = 0;
+
+    function handleHttpRequest() {
+      var timing = new Timing();
+
+      this.__inspector_timing__ = timing;
+      this.__inspector_ID__ = '' + lastRequestId++;
+      this.__inspector_documentURL__ = this._headers.host + this.path;
```
jkrems: This seems to be missing the protocol? Also, as mentioned in the other PR - this doesn't work when forging host headers or setting a port. Especially the latter part is fairly common when talking to internal services. Example:

```js
'use strict';

var http = require('http');

var deferToConnect = http.ClientRequest.prototype._deferToConnect;

http.ClientRequest.prototype._deferToConnect = function defer() {
  // This URL would load just fine
  console.log('Actually not %s', this._headers.host + this.path);
  return deferToConnect.apply(this, arguments);
};

var req = http.get({
  host: '10.255.255.1', port: 8080, headers: { Host: 'api.reddit.com' } // This hangs indefinitely
}, function(res) {});
```

lib/Injections/NetworkAgent.js
jkrems: > Correct me if I'm wrong here. And buffer.toString use utf8 by default.

You're totally right. Got that mixed up with something else.

jkrems: About the non-text: Maybe tracking the data as base64 (using `Buffer.concat`) and only converting it to a target encoding at the "last mile" could work better. I'm not even sure if some "exotic" text encodings wouldn't already break by forcing them to utf8 like this.

3y3: return value fixed

3y3:
> This seems to be missing the protocol? Also, as mentioned in the other PR - this doesn't work when forging host headers or setting a port. Especially the latter part is fairly common when talking to internal services.

fixed in `Network: find correct request url`

lib/Injections/NetworkAgent.js
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    if (!stack) stack = new Error().stack;
+
+    Error.prepareStackTrace = backup;
+
+    return stack.reduce(function(stack, frame) {
+      var fileName = frame.getFileName();
+
+      // v8NameToInspectorUrl
```
bajtos: Makes me wonder: is it possible to convert the filename to the url in NI process, so that we can reuse `convert.v8NameToInspectorUrl` and don't have to maintain the code in two places?

lib/Injections/NetworkAgent.js
```diff
@@ -0,0 +1,306 @@
...
+      if (/^\//.test(fileName)) {
+        fileName = 'file://' + fileName;
+      } else if (/^[a-zA-Z]:\\/.test(fileName)) {
+        fileName = 'file:///' + fileName.replace(/\\/g, '/');
+      } else if (/^\\\\/.test(fileName)) {
+        fileName = 'file://' + fileName.substring(2).replace(/\\/g, '/');
+      }
+
+      var url = fileName || frame.getEvalOrigin();
```
bajtos: What is the result of `getEvalOrigin()`, is it a front-end compatible URL or a fileName that must be converted to a URL first?

lib/Injections/NetworkAgent.js
```diff
@@ -0,0 +1,306 @@
...
+      var requestId = request.__inspector_ID__,
+        responseInfo = constructResponseInfo(response),
+        dataLength = 0;
+
+      debug.emitEvent('Network.responseReceived', responseInfo);
+      debug.emitEvent('Network._responseReceived', {
+        requestId: requestId
+      });
+
+      response.on('data', function(chunk) {
```
bajtos: IIRC, adding a `data` listeners changes the stream mode from Stream2/Stream3 back to Stream1. This may cause subtle and hard-to-identify bugs in the debugged application.

bajtos:
@3y3 Great stuff! I left few comments above.

lib/Injections/NetworkAgent.js
3y3: Converting in NI process looks ugly. But I can implement something like this in InjectorServer:
```js
var debug = require('v8-debug');
debug.convert = require(options['link-to-convert-script']);
```

On this line:
```js
fileName = debug.convert.v8NameToInspectorUrl(fileName)
```

lib/Injections/NetworkAgent.js
3y3: Completely agree. So, as a solution I can propose to pipe a request to a common transform stream and do all the things in `_transform` function. Will this be correct or I missed something?

lib/Injections/NetworkAgent.js
bajtos: The proposal in your comment looks good to me :+1:

lib/Injections/NetworkAgent.js
bajtos: To be honest, I don't know enough details about stream to tell what are the (possible) side effect of piping the response to another stream, but my gut feeling is that such solution is still not bullet proof.

I think you'll have to dig deeper and hook into an internal API. Below is a snippet of code that hooks into `res.push` and works for me in Node v0.10 and io.js v2.2.1:

```js
var http = require('http');

http.get('http://www.root.cz', function(res) {
  var push = res.push;
  res.push = function(chunk) {
    if (chunk) {
      // handle intercepted chunk of response data
      process.stderr.write(chunk.toString('utf-8'));
    }
    // call the original push method to preserve the observed behaviour
    return push.apply(this, arguments);
  };
  res.pipe(process.stdout);
});
```

The script prints the original response stream to stdout and the intercepted response stream to stderr. I ran it with `
node script.js 1>1.log 2>2.log` and called `diff -u 1.log 2.log` to verify that the output is the same.

3y3: Nice solution with push. It also prevent changing of stream mode to flowing. I'll use this solution.

3y3:
@bajtos , I fixed targetted issues. Unfortunately changes history was vwry dirty, so I prefer to squash all fixups.

What happened here:
- added `debug.convert`
- `response.push` used to handle `dataReceived` and `finishLoading` events.

@bajtos , @jkrems all your issues are fixed. It's ready to next round of review. First round was very useful.

jkrems:
Looking good! Two potential edge cases I found:

* Right now it ignores requests that are queued b/c of `maxSockets`, test case here: https://gist.github.com/jkrems/605a63f4ac068f42de30. This means that when a request times out (is aborted) because of sockets queueing, it will be invisible in the network tab
* Maybe bailing out if `options.socketPath` is present might be a good idea. Or adding a test making sure that the instrumentation doesn't break app code if the app is using file sockets somewhere.

bajtos:
LGTM. I think it's ok to squash most (all?) commits into a single one. Also please check why the CI build is failing, whether its related to the changes made in this PR.

I also think it's ok to address the two edge cases mentioned in @jkrems's comment in a new pull request, but feel free to fix them in this PR too. Whatever works best for you.

3y3:
Ok. I partially fixed issues from last @jkrems comments. I switched to new entry point for wrapper, and now I have some questions for discussion - look at code below.

From `_http_client.js` with my comments
```diff
function ClientRequest(options, cb) {
  var self = this;
  OutgoingMessage.call(self);

  if (typeof options === 'string') {
    options = url.parse(options);
  } else {
    options = util._extend({}, options);
  }

- NOTE 1:
+ var agent = options.agent;
+ var defaultAgent = options._defaultAgent || Agent.globalAgent;
+ if (agent === false) {
+  agent = new defaultAgent.constructor();
+ } else if ((agent === null || agent === undefined) &&
+           !options.createConnection) {
+   agent = defaultAgent;
+ }
+ self.agent = agent;
- Ok.  `request.agent` is undefined only if we specially define
- undocumented `_defaultAgent` option. Is it right or I missed something?
- So, are we need to handle this ugly case?

  var protocol = options.protocol || defaultAgent.protocol;
  var expectedProtocol = defaultAgent.protocol;
  if (self.agent && self.agent.protocol)
    expectedProtocol = self.agent.protocol;

  if (options.path && / /.test(options.path)) {
    // The actual regex is more like /[^A-Za-z0-9\-._~!$&'()*+,;=/:@]/
    // with an additional rule for ignoring percentage-escaped characters
    // but that's a) hard to capture in a regular expression that performs
    // well, and b) possibly too restrictive for real-world usage. That's
    // why it only scans for spaces because those are guaranteed to create
    // an invalid request.
    throw new TypeError('Request path contains unescaped characters.');
  } else if (protocol !== expectedProtocol) {
    throw new Error('Protocol "' + protocol + '" not supported. ' +
                    'Expected "' + expectedProtocol + '".');
  }

  const defaultPort = options.defaultPort ||
                      self.agent && self.agent.defaultPort;

  var port = options.port = options.port || defaultPort || 80;
- NOTE2:
+ var host = options.host = options.hostname || options.host || 'localhost';
- Are we need to display `options.hostname` instead of `options.host` if it exists?

  if (options.setHost === undefined) {
    var setHost = true;
  }

  self.socketPath = options.socketPath;

  var method = self.method = (options.method || 'GET').toUpperCase();
  self.path = options.path || '/';
  if (cb) {
    self.once('response', cb);
  }

  if (!Array.isArray(options.headers)) {
    if (options.headers) {
      var keys = Object.keys(options.headers);
      for (var i = 0, l = keys.length; i < l; i++) {
        var key = keys[i];
        self.setHeader(key, options.headers[key]);
      }
    }
    if (host && !this.getHeader('host') && setHost) {
      var hostHeader = host;
      if (port && +port !== defaultPort) {
        hostHeader += ':' + port;
      }
      this.setHeader('Host', hostHeader);
    }
  }

  if (options.auth && !this.getHeader('Authorization')) {
    //basic auth
    this.setHeader('Authorization', 'Basic ' +
                   new Buffer(options.auth).toString('base64'));
  }

  if (method === 'GET' ||
      method === 'HEAD' ||
      method === 'DELETE' ||
      method === 'OPTIONS' ||
      method === 'CONNECT') {
    self.useChunkedEncodingByDefault = false;
  } else {
    self.useChunkedEncodingByDefault = true;
  }

  if (Array.isArray(options.headers)) {
    self._storeHeader(self.method + ' ' + self.path + ' HTTP/1.1\r\n',
                      options.headers);
  } else if (self.getHeader('expect')) {
    self._storeHeader(self.method + ' ' + self.path + ' HTTP/1.1\r\n',
                      self._renderHeaders());
  }

  if (self.socketPath) {
    self._last = true;
    self.shouldKeepAlive = false;
    var conn = self.agent.createConnection({ path: self.socketPath });
    self.onSocket(conn);
  } else if (self.agent) {
    // If there is an agent we should default to Connection:keep-alive,
    // but only if the Agent will actually reuse the connection!
    // If it's not a keepAlive agent, and the maxSockets==Infinity, then
    // there's never a case where this socket will actually be reused
    if (!self.agent.keepAlive && !Number.isFinite(self.agent.maxSockets)) {
      self._last = true;
      self.shouldKeepAlive = false;
    } else {
      self._last = false;
      self.shouldKeepAlive = true;
    }
- NOTE 3:
- This is a new entry point for wrapping. Now we have full access to options so we can find `port` correctly.
- Also this prevent us from handling socketPath
- Side effects - also prevented handling of requests without agent,
- but it is a very strange behavior. (Look at NOTE 1)
- Can you describe me a situation where we need same requests?
+    self.agent.addRequest(self, options);
  } else {
    // No agent, default to Connection:close.
    self._last = true;
    self.shouldKeepAlive = false;
    if (options.createConnection) {
      var conn = options.createConnection(options);
    } else {
      debug('CLIENT use net.createConnection', options);
      var conn = net.createConnection(options);
    }
    self.onSocket(conn);
  }

  self._deferToConnect(null, null, function() {
    self._flush();
    self = null;
  });
}
```


@jkrems , about your gist example. 
```diff
+ agent: new http.Agent({
+        // -1 simulates an exhausted connection pool
+        maxSockets: -1
+      })
```
There you target me to situation when socket will be never initialised.
I think that current realisation (where we send `requestWillBeSent` after socket initialisation) is correct, otherwise we can't say what request really will be sent =)
All other cases from your gist are handled in new commits.

@bajtos , I fixed tests and squashed all changes to three commits:
- implementation
- backported changes
- tests

Please review `squash!` changes. There I fixed a completely missed `responseFailure` and checked behavior changing when we listen on `error` and `response` events.

3y3:
Sorry. I need to fix tests for 0.10.* but anyway it's ready to review.

bajtos:
@3y3 I took a quick look and did not find any obvious issues. Feel free to merge.

jkrems:
@3y3 The reason I included that as a test case is that it catches one of the nastier things to debug when working with a lot of service calls - resource limits + parallelization. E.g.:

* Using `maxSockets` to control the maximum number of concurrent sockets (blocked resources) your app will open against some upstream service ("being a good SOA neighbor" etc.)
* Combined with strict timeouts ("fail fast over hang long")
* Combined with parallel requests ("optimize the critical path")

Those three combined lead to situation where apps can sometimes start to make a request (from an app's perspective: the request was fired) but then before they even have a chance to acquire a socket, they time out and are aborted (are returning a `ETIMEDOUT` error from an app's perspective). Back when bugger's network tab was still working, these were the exact kinds of problems I used it for. Successful requests are relatively boring. ;)

3y3:
> Successful requests are relatively boring

Yes! :smile: , that's why I prefer to add `requestFailure` handling before merging.
So, now when we handle request error we will see in network tab
```
(failure)
ETIMEDOUT
```
or
```
(failure)
(unhandled) ETIMEDOUT
```

In other words all time when request was aborted or rejected we will handle it.
But there are two situations which we can't handle:
- User listen on `response` event but don't read the data.
- Socket is newer created for request (crazy `maxSockets = -1`). If I understand right this is impossible situation? (If `maxSockets > 0` then socket can be created from long time and we will see the problem in `timing` tab)

So, if I understand the problem you described will be handled in current implementation, or I missed something?

jkrems:
@3y3 The `maxSockets: -1` simulates a real condition. The `-1` just makes sure it happens on the very first request and you don't need 10 others to reproduce. Here's an example using a normal value for `maxSockets` that shows the same effect: https://gist.github.com/jkrems/605a63f4ac068f42de30#file-realistic-js

Output of the script:

```
> node req-test.js
Got socket for "/a"
Got socket for "/b"
Got socket for "/c"
Got socket for "/d"
Got socket for "/e"
Got socket for "/f"
Got socket for "/g"
Got socket for "/h"
Got socket for "/i"
Got socket for "/j"
[ 200,
  200,
  200,
  200,
  200,
  'ETIMEDOUT: /f',
  'ETIMEDOUT: /g',
  'ETIMEDOUT: /h',
  'ETIMEDOUT: /i',
  'ETIMEDOUT: /j',
  'ETIMEDOUT: /k',
  'ETIMEDOUT: /l',
  'ETIMEDOUT: /m',
  'ETIMEDOUT: /n',
  'ETIMEDOUT: /o' ]
```

As you can see it only fires the `socket` event for the first couple of requests (up to `j`), the remaining 5 never reach the listener.

3y3:
@jkrems , the output for your example (`throw err;` changed to `console.log(err);`):

output-network
output-console

Is this result acceptable, or I need fix something here?

3y3:
Merged. If there exists any bugs welcome with new prs.

PR #689, opened 2015-06-27
Implementation of HTTP Chrome endpoint

auchenberg:
This is first stab on an implementation of #684, that provides two new HTTP endpoints ```/json``` and ```/json/version```

``process.id`` is used as the unique identifier for the debugging target, as it's unique and consistent as long the process is running.v

PR #705, opened 2015-07-28
JSON endpoints /json and /json/list should return a JSON array

ackalker:
This enables interop with third-party debugging clients like
- chrome-repl from the chrome-remote-interface NodeJS module
- Kite and JsSlime for Emacs
- and more

In future this could be used to have multiple NodeJS processes
register with a single node-inspector process, which can then
advertise all these debuggable processes as "tabs" in the array.

This change improves on pull request #689, adding a /json/list
endpoint which should return the same data as the /json endpoint.

### DIFF
--- a/lib/debug-server.js
+++ b/lib/debug-server.js
@@ -38,7 +38,7 @@ function emptyJson(req, res) {
 }
 
 function jsonAction(req, res) {
-  res.json({
+  res.json([{
    'description': 'Node.js app (powered by node-inspector)',
    'devtoolsFrontendUrl': this.address().url,
    'id': process.pid,
@@ -46,7 +46,7 @@ function jsonAction(req, res) {
    'type': 'page',
    'url': '',
    'webSocketDebuggerUrl': this.wsAddress().url
-  });
+  }]);
 }
 
 function jsonVersionAction(req, res) {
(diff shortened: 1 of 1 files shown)


ackalker:
Here is an example using Kite for Emacs:

- In Emacs:
  - install the `kite` package from MELPA.
  - Evaluate `(require 'kite) (setq websocket-callback-debug-on-error nil)` to prevent Emacs from entering the debugger when Kite tries to execute some methods which have been removed from the debug protocol.
  - Create a file `test.js` with contents:
```js
console.log("start");
debugger;
console.log("end");
```

In a terminal:
  - In the directory containing `test.js`, run:
```sh
$ node-inspector &
$ node --debug --debug-brk test.js
```

In Emacs, start a Kite debug session:
  - Enter `C-u C-u M-x kite-console`
  - Accept `127.0.0.1` as Host, enter `8080` as Port, choose `node` as Tab

This should eventually (currently with still quite a few websocket errors) connect to node-inspector.
From here on, some of Kite's many features (including stack and object inspection) can be used.

Screenshot:
2015-07-28-025958_480x270_scrot-thumb

ackalker:
2015-07-28-031834_480x270_scrot-thumb

3y3:
@ackalker , thank you for contribution.
PR is ready to merge, but I need @auchenberg feedback.

@auchenberg , is this change sensible for you?

auchenberg:
Good catch, this should indeed be a list.

:+1:

3y3:
Landed as 4399f1e4623114fbe8111e618cb217ae24fc32e2

Thank you for contribution!

ackalker:
You're welcome :)

PR #750, opened 2015-09-20
Huge Startup Time Improvement!!

zanemcca:
I have added a new option `--ignore` that is somewhere in the middle of `--hidden` and `--no-preload`. Instead of hiding directories and files as `--hidden` does `--ignore` will avoid loading them all together. 

I saw the need for this because I had very slow startup times. Which lead me to `--no-preload`, but you cannot attach breakpoints to your files until they have been loaded. Which is less than ideal. 

I did not use `--hidden` because it uses RegExp and the ignore option in glob accepts glob syntax.
### Example - Ignore all node modules.
```
$ node-debug --ignore "**/node_modules/**" server/server.js
```
This makes startup load time insignificant while maintaining all meaningful debugging ability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,13 @@ $ node-inspector --hidden='["node_modules/framework"]'
 
 Note that the array items are interpreted as regular expressions.
 
+#### How do I ignore files so that they are not even loaded ?
+You can use glob notation with ```---ignore```. For example you could ignore all `node_modules` and the `client/` with the following: 
+
+```sh
+$ node-inspector --ignore "**/node_modules/**" --ignore "client/**"
+```
+
 #### UI doesn't load or doesn't work and refresh didn't help
 
 Make sure that you have adblock disabled as well as any other content blocking scripts and plugins.
@@ -169,7 +176,9 @@ When you are done cleaning up, hit refresh in the browser.
 
 #### Node Inspector takes a long time to start up.
 
-Try setting --no-preload to true. This option disables searching disk for *.js at startup.
+Try setting ```--ignore "**/node_modules/**"``` to avoid loading the node_modules/ directories.  
+
+If that does not work try setting --no-preload to true. This option disables searching disk for *.js at startup.
 Code will still be loaded into Node Inspector at runtime, as modules are required.
 
 #### How do I debug Mocha unit-tests?
@@ -310,6 +319,7 @@ so that sources earlier in this list override later ones.
 | **node-inspector**
 | --save-live-edit    |     | false   | Save live edit changes to disk (update the edited files).
 | --preload           |     | true    | Preload *.js files. You can disable this option<br/>to speed up the startup.
+| --ignore            |     | []      | Array of files to ignore from loading. You can use this to speed up the startup while maintaining the ability to set breakpoints.
 | --inject            |     | true    | Enable injection of debugger extensions into the debugged process. It's possible disable only part of injections using subkeys `--no-inject.network`. Allowed keys : `network`, `profiles`, `console`.
 | --hidden            |     | []      | Array of files to hide from the UI,<br/>breakpoints in these files will be ignored.<br/>All paths are interpreted as regular expressions.
 | --stack-trace-limit |     | 50      | Number of stack frames to show on a breakpoint.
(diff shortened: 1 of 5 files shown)


zanemcca:
Could I get some feedback on this?

DylanPiercey:
@zanemcca would this speed up load times?

zanemcca:
For me it significantly improved load times.
Basically it allows you to completely ignore files and folder. So I ignore node_modules (~ 1Gb) since I rarely need to debug anything in there and it now loads within 5 seconds.

zanemcca:
@3y3 Can I please have some feedback?

grsabreu:
This seems interesting!
Any feedback in here?

joeyrobert:
I've tested this with [a project of mine](https://bitbucket.org/joeyrobert/ceruleanjs/branch/fix/faster-debug#diff). It brings the startup time for `node-debug` down to 8.56s from 21.03s without this fix. The size of `node_modules` folder side is 72.4 MB / 8,658 files. This test was run on Windows 10 with a fast SSD.

I only have one `dependencies` in my project. The majority of my `node_modules` is `devDependencies`, so debugging `node_modules` code is not something I need to do. This is one anecdotal benchmark, but I'd say the `ignore` option improves startup time significantly for those that opt into it.

zanemcca:
Thanks for the test @joeyrobert

3y3:
Wow! I'm sorry for ignoring this pr for long time! I missed all notifications...
Mmm... I think that I worked on the same problem in `next` branch.
Unfortunately I can't approve this pr, because DevTools uses RegExps everywhere and in future `hidden` option will be part of `framework blackboxing` feature.
In `next` branch I completely avoid using of `glob` for collecting resources tree. I'll backport this changes to current version (it is latest which supports node < v4.0) after `Node Inspector Next` release.

Or maybe someone from this branch can help me to backport new algorithm to current version (with ugly async syntax).

I'm sorry again that I missed this pr - I have a lot of notifications.

zanemcca:
@3y3 I am not sure what you are looking for. 
Are you asking for help to move the changes into RegEx for version `v0.12.5`?

Also when you say you have done something similar in the `next` branch does that mean that the ported changes would be deprecated out after `next`'s release?

Also how have you implemented it in `next`? Is it done through the `hidden` or `--no-preload` or another mechanism entirely?

3y3:
> Also when you say you have done something similar in the next branch does that mean that the ported changes would be deprecated out after next's release?

No, I'll continue to contribute in 0.x node-inspector branch for this year. I will backport more sensible fixes, but no features.
All features will be implemented in 1.0.0-(alfa -> beta -> stable)

> Also how have you implemented it in next? Is it done through the hidden or --no-preload or another mechanism entirely?

This is a part of `hidden` option now. There is no true reason to load files if we can't pause in their source. Also current logic prevents step into folders matched by hidden RegExps.

https://github.com/node-inspector/node-inspector/blob/next/lib/ScriptFileStorage.js#L224-L255

3y3:
Take a chance to current node-inspector master

zanemcca:
I can confirm that 8dd50abcebe6d0029a1d48d74bccacf92052f1bf has solved this problem.

#### Before - with or without `--hidden`
- Appx. 2 min to load

#### After - without `--hidden`
- 15 sec load time

#### After - with `--hidden node_modules/`
- 6 sec load time

zanemcca:
I'm not sure if you did it on purpose but you rolled package.json back to v0.12.3?

3y3:
> I'm not sure if you did it on purpose but you rolled package.json back to v0.12.3?

Ouch!!! Thank you for targetting!

> I can confirm that 8dd50ab has solved this problem.

Nice! So I will prepare v0.12.6 soon

3y3:
@zanemcca , do you see any issues in tree? Maybe I ignore some files which shoul not to be ignored?

zanemcca:
@3y3 I do not see anything missing in the tree.

You do however have some debugs that still refer to `glob` as well as including it in package.json and in  a test.
Also when you rolled package.json back you also rolled back the version of a couple of dependencies.

3y3:
Released 0.12.6

PR #772, opened 2015-11-06
New option to exclude preload files and rewriting ScriptFileStorage.listScripts

shitiven:
I have added a new option `--preload-exclude`.  Sometimes I don't want  `node_modules` to preload, because that files all preload together will cause my debug efficiency.
#### Example - preload exclude all node_modules.
```bash
node-inspector  --preload-exclude .*\\/node_modules\\/.*
```
But I found that even though I blocked these files, still very slow startup times.   I rewrote `listScripts` in ScriptFileStorage.js.  

Now It can be very fast startup times, and even if  did not use `--preload-exclude .*\\/node_modules\\/.*`

### DIFF
--- a/lib/ScriptFileStorage.js
+++ b/lib/ScriptFileStorage.js
@@ -28,6 +28,7 @@ function ScriptFileStorage(config, session) {
   config = config || {};
   this._scriptManager = session.scriptManager;
   this._noPreload = config.preload === false;
+  this._preloadExclude = config.preloadExclude||[];
 }
 
 var $class = ScriptFileStorage.prototype;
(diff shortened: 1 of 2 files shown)


zanemcca:
@shitiven I added an [option](#750) you may want to look at.  
It adds an option that will ignore the directories and files that you pass in using glob notation.
Example. 
```node-debug --ignore "**/node_modules/**" <filename>```

I use it to ignore node_modules which makes my debug startup time go from close to a minute down to a few seconds.

3y3:
Thank you for contribution!
In node-inspector 0.12.6 released alternative solution which solves this problem

PR #789, opened 2015-12-09
Minor Plugin API Extensions

eplawless:
#### Overview

Hello! I'd like to make a few minor changes to the plugin API which will make it easier to keep the vanilla node-inspector source tree untouched (and up to date) and ship plugins alongside.

I'm using node-inspector plus some custom plugins to provide JS debugging support for the Netflix application on TVs and game consoles.

This change would make it so I can stop shipping a forked version :)

#### Proposed Changes
* Added --plugin-path argument for specifying root plugin path when plugins are enabled.
* Added plugin/manifest.json "override" property to allow finer-grained control of inspector.json.

### DIFF
--- a/lib/debug-server.js
+++ b/lib/debug-server.js
@@ -14,7 +14,6 @@ var http = require('http'),
     plugins = require('./plugins'),
     InspectorJson = plugins.InspectorJson,
     ProtocolJson = plugins.ProtocolJson,
-    PLUGINS = plugins.cwd,
     OVERRIDES = path.join(__dirname, '../front-end-node'),
     WEBROOT = path.join(__dirname, '../front-end');
 
@@ -85,8 +84,13 @@ inherits(DebugServer, EventEmitter);
 DebugServer.prototype.start = function(options) {
   this._config = extend({}, options);
   this._isHTTPS = this._config.sslKey && this._config.sslCert ? true : false;
+
+  plugins.init(this._config);
+
   this._inspectorJson = new InspectorJson(this._config);
   this._protocolJson = new ProtocolJson(this._config);
+
+  var PLUGINS = plugins.CWD;
 
   var app = express();
   var httpServer;
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -2,8 +2,11 @@ var fs = require('fs'),
     inherits = require('util').inherits,
     path = require('path');
 
-var CWD = path.join(__dirname, '../plugins'),
-    THROW_CONFLICTS = true;
+var THROW_CONFLICTS = true;
+
+function getPluginPath(config) {
+  return (config && config.pluginPath) || path.join(__dirname, '../plugins');
+}
 
 function PluginError(message) {
   this.name = 'Plugin Error';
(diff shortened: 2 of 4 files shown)


eplawless:
@3y3 Wondering if I could get your feedback on this change, whether it's missing anything etc.

eplawless:
The reason for the "override" property in addition to "exclude" is in order to re-enable or tweak features that node-inspector has already set.

My immediate use case (likely with more to follow) is this plugin:

**manifest.json**
```json
{
  "name": "restore-elements-panel",
  "override": [{ "name": "elements", "type": "autostart" }]
}
```

**module.json**
```json
{}
```

3y3:
In current time I'm ready to accept this changes, but keep in mind, that plugins api is unofficial because it is uncompleted:
- [ ] DevTools Client
- [ ] Inspector process
- [ ] Application process
- [ ] Plugins installer

I will try to save ability to setup plugins dir, but I'm not sure, that there will not problems with `Plugins installer` step.
So I'll notify you about this as a first user of plugins api =)
And I'll call you in pull request for review.

I have some inline questions ip pr

.gitignore
```diff
@@ -6,3 +6,5 @@ v8.log
 /test/fixtures/work
 npm-debug.log
 node-inspector-*.tgz
+*.swp
+*.swo
```
3y3: Unrelated change

lib/plugins.js
```diff
@@ -2,8 +2,11 @@ var fs = require('fs'),
     inherits = require('util').inherits,
     path = require('path');
 
-var CWD = path.join(__dirname, '../plugins'),
-    THROW_CONFLICTS = true;
+var THROW_CONFLICTS = true;
+
+function getPluginPath(config) {
```
3y3: Is it possible to save `CWD` option and rewrite in init function:

```js
var plugins = require('./plugins.js'); // ready to work with default parameters
plugins.init(options); // override CWD
```
?

lib/plugins.js
```diff
@@ -30,18 +33,35 @@ function mergeByName(acceptor, donor, name, onConflict) {
...
-  dirlist = [];
+function _setMockPlugins(pluginPath, plugins) {
+  cachedPlugins[pluginPath] = plugins;
 }
-  
-dirlist.reduce(function(plugins, subdir) {
-    var _path = path.resolve(CWD, subdir, 'manifest.json'),
-        manifest;
+
+function getPlugins(config, ignoreCache) {
```
3y3: There we use `ignoreCache ` with `ignoreCache => true`?

lib/debug-server.js
```diff
@@ -119,7 +119,7 @@ DebugServer.prototype.start = function(options) {
   app.get('/', debugAction.bind(this));
   app.get('/debug', redirectToRoot.bind(this));
   app.use('/node', express.static(OVERRIDES));
-  app.use('/plugins', express.static(PLUGINS));
+  app.use('/plugins', express.static(getPluginPath(options)));
```
3y3: Why we need to recalculate plugins on each request to them?

lib/plugins.js
eplawless: Yes, you could call `getPlugins({ plugins: true, pluginPath: '/foo/bar' }, true)` and it will ignore the cached version and read from the directory again.

I added a separate argument so people don't have to extend or mutate the config object.

3y3:
Please isolate `override` option in new commit in current pr.

lib/plugins.js
eplawless: Yeah, I could definitely make that change if you like.

.gitignore
eplawless: I'll remove it from the PR.

eplawless:
Sounds good, I'll make those changes.

lib/plugins.js
3y3: Yes, I understand how it can be used, but I don't see where we use this logic.

I'm not sure that reinitialisation of plugins on the fly is good think. It's useful in plugins development, but for end user this can cause some unexpected problems. Can we avoid caching this logic on firs step?

eplawless: I can definitely remove it, it's unused currently.

eplawless:
Made those edits, let me know if I missed anything?

lib/plugins.js
eplawless: It's used in lib/plugins.js now, but not in test/plugins.js; I can modify the test file as well, if you like.

eplawless:
We're going to be actively developing and maintaining plugins, so I'm perfectly happy to adapt to the plugins API as it evolves and changes.  I think it's going to enable some really neat tools for our teams.

If you need some help with node-inspector itself at any point, I think I can swing some dedicated time for contributing to it.

eplawless:
@3y3 is there anything else I should change?

3y3:
No. All ready and merged. I need to fix some bugs before next release, so wait next release after weekend.

Thank you.

PR #798, opened 2015-12-18
Fix node-inspector redirecting with invalid ws url if no port is spec…

cainjonm:
…ified


3y3:
Thank you for contribution!

PR #839, opened 2016-03-08
Don't fail when there's a loop in symlinks

Y--:


3y3:
Thank you for contribution.
Can you please add test here?

Y--:
Hi @3y3, thanks for considering my PR.
Sorry for the delay, I've finally got time to add a test.
It is not ideal but I could not find a better way to check whether an error occurred here or not :
https://github.com/node-inspector/node-inspector/blob/8dd50abcebe6d0029a1d48d74bccacf92052f1bf/lib/ScriptFileStorage.js#L246

Let me know if you have any suggesting to improve it.

Thanks!

3y3:
This is not important to check errors here.
You need only check that you have expected result with correct input.

Y--:
The expected result is tested.
Checking the errors that way is the only way I found to ensure that the algorithm won't follow the symbolic links until it faces an "ELOOP" error.
I can modify the implementation to have it being tested properly though
On my environment, I simply cannot use the current version of node-inspector because of this behavior, since it takes several minutes to start (or crash).

Issue #767, opened 2015-10-19
Cannot install on windows with node 4.1.1


togakangaroo:
Just tried to uninstall and re-install node-inspector and it just won't work.

[Here's the npm debug log](https://www.dropbox.com/s/to54hili33kco2b/npm-debug.1.log?dl=0)

Please note that while the actual issue is being thrown by bufferutil, [according to them](https://github.com/websockets/bufferutil/issues/22) it is because the dependency is out of date.

```
W:\> npm uninstall -g node-inspector
W:\> npm install -g node-inspector
npm WARN engine xmlbuilder@2.2.1: wanted: {"node":"0.8.x || 0.10.x"} (current: {"node":"4.1.1","npm":"3.3.4"})
npm WARN install Couldn't install optional dependency: Unsupported

> v8-debug@0.5.4 preinstall C:\Users\George\AppData\Roaming\npm\node_modules\.staging\v8-debug-2835e2319fe238f028ae8d5d2df32c30
>


> v8-profiler@5.3.2 preinstall C:\Users\George\AppData\Roaming\npm\node_modules\.staging\v8-profiler-8ba8d2a158f95e40ca38eab5fe19a454
>

C:\Users\George\AppData\Roaming\npm\node-inspector -> C:\Users\George\AppData\Roaming\npm\node_modules\node-inspector\bin\inspector.js
C:\Users\George\AppData\Roaming\npm\node-debug -> C:\Users\George\AppData\Roaming\npm\node_modules\node-inspector\bin\node-debug.js

> bufferutil@1.2.1 install C:\Users\George\AppData\Roaming\npm\node_modules\node-inspector\node_modules\bufferutil
> node-gyp rebuild


C:\Users\George\AppData\Roaming\npm\node_modules\node-inspector\node_modules\bufferutil>if not defined npm_config_node_gyp (node "C:\Users\George\AppData\Roaming\npm\node_modules\npm\bin\node-gyp-bin\\..\..\node_modules\node-gyp\bin\node-gyp.js" rebuild )  else (node  rebuild )
Building the projects in this solution one at a time. To enable parallel build, please add the "/m" switch.
TRACKER : error TRK0005: Failed to locate: "CL.exe". The system cannot find the file specified. [C:\Users\George\AppData\Roaming\npm\node_modules\node-inspector\node_modules\bufferutil\build\bufferutil.vcxproj]


gyp ERR! build error
[...]

PR #848, opened 2016-03-29
Bump ws dependency to 1.0.1 (eliminates dependency on bufferutil)

asselin:
This fixes issue #767.  Starting with version 1.0.0, ws no longer depends on bufferutil, eliminating the need to install a compiler on Windows and the whole node-gyp thing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "v8-debug": "~0.7.1",
     "v8-profiler": "~5.6.0",
     "which": "^1.1.1",
-    "ws": "^0.8.0",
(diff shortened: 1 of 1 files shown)


asselin:
I'm not sure why the 2 Travis builds failed... The failing testcase doesn't fail on my Windows or Mac system (although there is another testcase that fails on Windows due to / vs \ in expected path names).

3y3:
Landed as c769ae92bfec5dc8fcb4d291abf94271b17dabfa
Thank you for contribution!

Issue #832, opened 2016-02-28
node-inspector and babel-register


mdotasia:
I am using this setup https://babeljs.io/docs/setup/#mocha to run mocha tests written in ES6. I would like to debug a mocha test with node-inspector. Running ES6 mocha tests works fine with this setup.

In order to debug, I run `node-inspector` in one terminal and

```
mocha --compilers js:babel-register --debug-brk
```
 in another. I can debug in principle, i.e. on any ES5 code I can set break points. Any breakpoints in my ES6 tests get ignored. It looks like node-inspector does not pick up any sourcemaps.

Now babel-register does create sourcemaps by default, only I have no idea whether or not it saves compiles files to disk and whether node-inspector has a chance to get its hand on the sourcemaps.

Has anyone managed to debug code in node-inspector that has been processed with babel-register?

mdotasia:
I did some more analysis. It turns out that on Windows 7 with Node 5.6.0 I can debug mocha tests without problem. When I debug a `test.js` (ES6) I launch node-inspector/mocha as described in my previous comment in Chrome. Then I can set breakpoints in `test.js`. I press continue and then I see that node-inspector replaces `test.js` with the compiled versions and a `test.js.source` appears, which contains the original source ES6 code. In `test.js.source` I can set additional breakpoints and all is good.

Now I switch to my Mac and do exactly the same thing. In Chrome the test exits with an error. In Opera I have more luck and I can launch the debug session and set a breakpoint in `test.js`. I press continue and the debugger stops in the compiled file. The original ES6 file does not seem to be linked to the compiled file, even though I see inline sourcemaps. None of the magic that happened in Windows seems to happen on my Mac.

What's going wrong here?

3y3:
Hm... Last time then I fix source maps I make if for Win.
Maybe now I need to inspect problems for Mac.
I don't promise that this will be fast - in current time I intentionally work on major release.

Is your issue happens only for main file or also for all sub scripts?

mdotasia:
In above scenario I'm running my tests with

```
mocha --recursive --compilers js:babel-register
```

and then I'm trying to debug only the test that failed. Weirdly on Windows it work all fine. But Mac seems to be broken.

mdotasia:
I played around with `node-inspector` debugging a bit more and tested debugging with normal babel compilation, i.e. not using `babel-register` for on-the-fly compilation but **compile first** and **then debug**. Here is what I found for debugging with `node-inspector`:

[...]

PR #849, opened 2016-03-30
Fix issue #832, "node inspector and babel-register".

apomortsev:
This fixes the issue on Mac OSX where it appears to be impossible to
enable any breakpoints while debugging ES6 code dynamically transpiled
by either babel-node or babel-register.
The initial cause of the problem is that the source and the target file
names in source maps produced by babel-register are exactly equal and
have no paths.
node-inspector already contains code that addresses this problem:
_checkSourceMapIssues()/fixWrongFileName() in lib/ScriptManager.js,
which appends ".source" extension to the source file name in the source
map, when name duplication is detected. Unfortunately the code converting
URLs to absolute paths in that method does not work correctly with
Linux/OSX URLs: it strips the root slash from the path thus rendering
path invalid. This commit fixes this bug and also, exactly same bug
in lib/ScriptManager.normalizeName().

### DIFF
--- a/lib/ScriptManager.js
+++ b/lib/ScriptManager.js
@@ -241,7 +241,10 @@ ScriptManager.prototype = Object.create(events.EventEmitter.prototype, {
 
   _checkSourceMapIssues: {
     value: function(inspectorScriptData, sourceMap) {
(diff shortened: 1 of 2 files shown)


3y3:
Thank you for contribution!
Source maps is a big problem of Node Inspector...

PR #861, opened 2016-04-27
bump v8-profiler version as to be compatible with node

RobertWHurst:
This bumbs the version of v8-profiler so it works on Node v6. An issue is here for reference => https://github.com/node-inspector/v8-profiler/issues/89


RobertWHurst:
WIP: I believe I will need to bump the version of v8-debug also before this is ready.

RobertWHurst:
Blocked by https://github.com/node-inspector/v8-debug/issues/25

3y3:
At current time publishing of completely working version is blocked by node-pre-gyp, which know nothing about node 6.0.0 - this disallow to save prebuilt binaries for 6.0.0 in to the package.

But v8-debug 0.7.3 is ready to work with node 6.0.0. And current node-inspector version is also ready (without any bumps) is you have correct build environment (node-gyp requirements)

So, I'll bump package versions some later, after node-pre-gyp update.

RobertWHurst:
@3y3 Okie dokie. Should I close this for now?

PR #889, opened 2016-06-17
add support for jsx, es6, and coffee file extensions

mattkime:
this needs tests and perhaps some configuration support but i wanted to get the ball rolling by creating a (flawed) pr.


mattkime:
this apparently affects the loading of non-sourcemaped files. (only see main file under file://, not full file set  although webpack:// has everything...yeah, i'm using webpack)

PR #890, opened 2016-06-20
doc: Troubleshooting long startup time / UI not loading

jakub-g:
Hello,

I've noticed that sometimes Node Inspector wouldn't load for me in some repos, and I didn't know why. Finally today I spent more time and tracked the issue, and copying stuff to a separate folder made wonders. I think some people my find my advice useful :)

Regards
Jakub


jakub-g:
@3y3 what do you think about this PR? it took me *months* of node-inspector seemingly randomly not working for some projects to realize what's going on. I think this piece of documentation can prevent many people from pulling their hair.

PR #914, opened 2016-09-01
Fix callback call in InjectorClient._findNMInScope

Kreozot:
When error is defined (in this case it means that NM is doen't have any elements) callback calling will be throwing an exception. Because in this case NM[0] is not defined. So we need to check the error variable.

### DIFF
--- a/lib/InjectorClient.js
+++ b/lib/InjectorClient.js
@@ -108,7 +108,7 @@ InjectorClient.prototype._findNMInScope = function(funcHandle, cb) {
       if (!NM.length)
         error = new Error('No NativeModule in target scope');
 
-      cb(error, NM[0].ref);
+      cb(error, error ? null : NM[0].ref);
     }.bind(this));
 };
 


aknuds1:
I noticed this bug too, as I'm experiencing the error case.

braydonf:
Works for me, doesn't crash when trying to attach.

vinceyuan:
https://github.com/facebook/nuclide/issues/673#issuecomment-270058908 mentioned this bug which causes failure of debugging React-Native app.

Please approve and merge.

vinceyuan:
@Kreozot I think we need to check if NM.length==0.
`cb(error, NM.length==0 ? null : NM[0].ref);`

This bug prevents me from debugging React Native project. Hope it will be merged soon and all react native developers will benefit.

vinceyuan:
@Kreozot Please take a look at my previous comment.

Kreozot:
@vinceyuan I don't see any difference between checking `error` (which defines when NM.length equals zero) and checking `NM.length` itself. Do you think that checking `NM.length` is more clear way?

Kreozot:
By the way, @3y3, @bajtos, @dannycoates, is the project is abandoned? Any alternatives/forks? Any plans to transfer maintenance?

vinceyuan:
@Kreozot Yes. I think so. But you are right. Actually there is no difference in this context.

Kreozot:
Let's wait what the maintainers will say :)

bajtos:
Hello, while I am no longer actively involved with Node Inspector and thus I am not familiar with the current codebase, the fix looks reasonably to me. I have landed the patch.

> is the project is abandoned? Any alternatives/forks? 

In my opinion, the built-in `node --inspect` command is far more superior than `node-inspector` can ever become, I have personally switched long time ago. I believe the built-in inspector supports profiling too.

> Any plans to transfer maintenance?

I cannot speak for @3y3, but if there are volunteers, then I am myself happy to hand the maintenance over to them.

PR #938, opened 2016-10-19
patched InjectorClient.js for compatibility above nodejs version 6.3.1

MasterJames:
Despite the console log "Error: No NativeModule in target scope", which still needs addressing, this patch does enable it to work, with latest (nodejs 6.9.0)
many others to recognize for this @Virtual-Machine @asi-jmasson @dbreese


MasterJames:
Okay I thought there must be something else people are using now. That makes sense, thanks.
`node --inspect your.js`
seems to be integrated now.

PR #939, opened 2016-10-27
Add example to connect to remote server to run debug

harrissAvalon:
I don't know if this is out of scope, but I thought it would be helpful to have an example of how to connect to a remote server to do some debugging.


hustxiaoc:
Please see the document here. https://github.com/node-inspector/node-inspector/pull/1002

Issue #903, opened 2016-08-11
enter key in console and "add watch expression" performs line-break not execute


NilsFrkal:
When entering commands/variables in the console (both break-point and non-break-point scenarios) or when specifying a variable under "add watch expression" the "enter" key does not execute the command/variable, but does a line-break instead. I could not find any other way to actually execute the command in the console. How can this behavior be changed or is this a bug? Thank you very much.

node-inspector version 0.12.8
nodejs 5.12.0
chrome Version 54.0.2824.0 dev-m (64-bit)

corsen2000:
Having the same issue. Any advice?

NilsFrkal:
I did not work out how to fix it. Interestingly however, I had this issue on both my desktop and my laptop. Because of an issue I had to revert back to an older backup for my desktop (this backup did not have node inspector and chrome installed). After re-installing above sws the issue did not occur anymore whilst on the laptop it still exists. If you have time try de-installing both chrome and node-inspector and re-install.

ac9:
I ran into this same issue. I could not fix it by reinstalling chrome and node-inspector.
I could only fix it by rolling back to Version 51.0.2704.106 (64-bit). 

node-inspector version 0.12.8
nodejs v4.2.2
Chrome 54.0.2840.59
OS X El Capitan Version 10.11.6

elliotatRetireup:
I'm not sure if its related, but I have the same issue as of yesterday. I didn't change anything I was doing with inspector. Can't fix by restarting anything. 

Since this issue showed up, I have noticed inconsistent behavior with the inspector. For instance, getting an 'is not a function' error for using .toLowerCase() on a string... only sometimes though.

Then I noticed the same thing happening with console.log(). I have no clue why this is happening but would love to be able to fix this

newpost:
i have same issue

pyalxx:
Same issue here

OS: macOS Sierra 10.12.1
Chrome: Version 54.0.2840.71 (64-bit)
node-inpector: 0.12.8

Note: same issue in watch - enter starts a new line

ejhayes:
It looks like this issue might be related to a Chrome update in [54.0.2840.71](https://googlechromereleases.blogspot.com/2016/10/stable-channel-update-for-desktop.html).  I rolled back to `51.0.2704.106` (https://google-chrome.en.uptodown.com/mac/old, the link to `53.0.2785.101` was broken) which resolved the issue.

- Node: 4.2.6
[...]

PR #942, opened 2016-11-01
fix(front-end): DOMExtension, isEnterKey

iamfrontender:
Seems that a lot of time passed since last front-end synchronisation with ChromeDevTools.
Looks like full upgrade would be a proper fix.

### DIFF
--- a/front-end/platform/DOMExtension.js
+++ b/front-end/platform/DOMExtension.js
@@ -778,7 +778,7 @@ Document.prototype.deepElementFromPoint = function(x, y)
  */
 function isEnterKey(event) {
     // Check if in IME.
-    return event.keyCode !== 229 && event.keyIdentifier === "Enter";
+    return event.keyCode !== 229 && ((event.keyIdentifier || event.key) === "Enter");
 }
 
 function consumeEvent(e)


iamfrontender:
Fixes #903 though

iamfrontender:
It's not a proper fix. More of a workaround. 
Proper fix would be to integrate with latest ChromeDevTools frontend. It'd fix broken F8, arrows and the rest of issues people are having after upgrading from Chrome 51.

ryanspillsbury90:
This is a big headache for a lot of people. Please consider merging this in and doing the proper fix later on. Thanks

iamfrontender:
@ryanspillsbury90 , just to prevent you from wasting time. This one is unlikely to be merged, repo doesn't look maintained.

Mido22:
Yeah, just checked the last commit seems to be on April 9 :(

cgadam:
@iamfrontender I saw that your workaround solves the problem of the Enter not working in the console, however, if I assign something new to an already existing variable the value does not get reflected. Do you know why that might happen?

This same thing does not happens if you define the variable in the console, but it does happens for existing ones:

Variables defined in console

<pre>
var aVar = 2
undefined
aVar
2
aVar = 3
3
aVar
3
<pre>

versus

Variables already defined in the code and trying to overwrite its value during a breakpoint:
<pre>
ip
"1.1.1.1"
ip='Hello'
"Hello"
ip
"1.1.1.1"
<pre>

a7madgamal:
any updates?

mmrko:
Just to add to the discussion above, I'm using the following `sed` statement to patch the Enter key support for the time being:

```bash
sed -i 's/return event.keyCode !== 229 && event.keyIdentifier === "Enter";/return (event.keyCode !== 229 \&\& event.keyIdentifier === "Enter") || event.keyCode === 13;/' node_modules/node-inspector/front-end/platform/DOMExtension.js
```

codeHatcher:
I'm waiting for this PR as well. I think an entire community is actually

a7madgamal:
I hope someone updates this great tool, it's missing A LOT from the current devtools 😿

hustxiaoc:
duplicate  https://github.com/node-inspector/node-inspector/pull/955

PR #955, opened 2016-12-07
KeyboardEvent.key instead of KeyboardEvent.keyIdentifier  in front-end 

sparkleholic:
As `KeyboardEvent.keyIdentifier` is removed from the recent browsers 
including chrome 54 rev, front-end should use `KeyboardEvent.key` instead.

(note.)
[https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyIdentifier](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyIdentifier)

### DIFF
--- a/front-end/platform/DOMExtension.js
+++ b/front-end/platform/DOMExtension.js
@@ -778,7 +778,7 @@ Document.prototype.deepElementFromPoint = function(x, y)
  */
 function isEnterKey(event) {
     // Check if in IME.
-    return event.keyCode !== 229 && event.keyIdentifier === "Enter";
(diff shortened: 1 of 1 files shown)


gabmontes:
This PR fixes #903. Any news on test/merge/release? Thanks!!

codeHatcher:
Is it possible to merge this, it looks like the tests failed because of a timeout. There is a dependency on this package which has been causing issues for months over here https://github.com/meteor/meteor/issues/8469

damonomad:
I'm also excited for this to be merged, this fix saved me a LOT of time.

PR #956, opened 2016-12-09
[0.7.x] Change node-inspector to support the remote debugging.

sparkleholic:
### :Detailed Notes:
node-inspector had a restriction to be run on the same
host machine with the debuggable nodejs process.
This patch to release this limitation to be able to debug even though
the nodejs process is running on the different machine.

### :Example:
`./bin/inspector.js --debug-host='192.168.0.12' --debug-port=5858`
or put the host address and port in the config.json and run inspector.js
or can change the value of the host argument in the url.
  ex) http://127.0.0.1:8080/debug?host=192.168.0.12&port=5858

### :Etc:
- This PR aim to merge into node-inspector v0.7.x branch
- As `KeyboardEvent.keyIdentifier` is removed from the recent browsers including chrome 54 rev,
  front-end was updated to use KeyboardEvent.key instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,7 @@ Command line options:
 ```
 --debug-brk, -b         Break on the first line (`node --debug-brk`) [default: true]
 --web-port, -p, --port  Node Inspector port (`node-inspector --web-port={port}`)
+--debug-port, -d        Node/V8 debugger ip address (`node-inspector --debug-host={port}`)
 --debug-port, -d        Node/V8 debugger port (`node --debug={port}`)
 --cli, -c               CLI mode, do not open browser.
 --version, -v           Print Node Inspector's version.
@@ -291,6 +292,7 @@ Use dashed option names in RC files. Sample config file:
 {
   "web-port": 8088,
   "web-host": null,
+  "debug-host": "127.0.0.1",
   "debug-port": 5858,
   "save-live-edit": true,
   "no-preload": true,
@@ -305,6 +307,7 @@ List of predefined options:
 --help               |             | Print information about options
 --web-port           |    8080     | Port to host the inspector
 --web-host           |  127.0.0.1  | Host to listen on
+--debug-port         |  127.0.0.1  | Host where the debugging app runs on
 --debug-port         |    5858     | Port to connect to the debugging app
 --save-live-edit     |    false    | Save live edit changes to disk
                      |             |   (update the edited files)
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ Command line options:
 ```
 --debug-brk, -b         Break on the first line (`node --debug-brk`) [default: true]
 --web-port, -p, --port  Node Inspector port (`node-inspector --web-port={port}`)
---debug-port, -d        Node/V8 debugger ip address (`node-inspector --debug-host={port}`)
+--debug-host            Host where the debugged app is running (`node-inspector --debug-host={port}`)
 --debug-port, -d        Node/V8 debugger port (`node --debug={port}`)
 --cli, -c               CLI mode, do not open browser.
 --version, -v           Print Node Inspector's version.
@@ -292,8 +292,8 @@ Use dashed option names in RC files. Sample config file:
 {
   "web-port": 8088,
   "web-host": null,
-  "debug-host": "127.0.0.1",
   "debug-port": 5858,
+  "debug-host": null,
   "save-live-edit": true,
   "no-preload": true,
   "hidden": [],
@@ -307,8 +307,8 @@ List of predefined options:
 --help               |             | Print information about options
 --web-port           |    8080     | Port to host the inspector
 --web-host           |  127.0.0.1  | Host to listen on
---debug-port         |  127.0.0.1  | Host where the debugging app runs on
 --debug-port         |    5858     | Port to connect to the debugging app
+--debug-host         |  127.0.0.1  | Host where the debugged app is running
 --save-live-edit     |    false    | Save live edit changes to disk
                      |             |   (update the edited files)
 --preload            |    true     | Preload *.js files. You can disable this option
--- a/README.md
+++ b/README.md
@@ -257,7 +257,6 @@ Command line options:
 ```
 --debug-brk, -b         Break on the first line (`node --debug-brk`) [default: true]
 --web-port, -p, --port  Node Inspector port (`node-inspector --web-port={port}`)
---debug-host            Host where the debugged app is running (`node-inspector --debug-host={port}`)
 --debug-port, -d        Node/V8 debugger port (`node --debug={port}`)
 --cli, -c               CLI mode, do not open browser.
 --version, -v           Print Node Inspector's version.
--- a/bin/node-debug.js
+++ b/bin/node-debug.js
@@ -27,10 +27,6 @@ var argvOptions = {
     type: 'number',
     description: 'Node/V8 debugger port (`node --debug={port}`)'
   },
-  'debug-host': {
-    type: 'string',
-    description: 'Host where the debugged app is running (`node-inspector --debug-host={url}`)'
-  },
   nodejs: {
     type: 'string',
     description: 'Pass NodeJS options to debugged process (`node --option={value}`)\n' +
@@ -155,10 +151,8 @@ function parseArgs(argv) {
   }
 
   var inspectorPort = options['web-port'] || 8080;
-  var debugHost = options['debug-host'] || '';
   var inspectorArgs = extractPassThroughArgs(options, argvOptions)
-    .concat(['--web-port=' + inspectorPort])
-    .concat(['--debug-host=' + debugHost]);
+    .concat(['--web-port=' + inspectorPort]);
 
   return {
     printScript: printScript,
@@ -171,7 +165,6 @@ function parseArgs(argv) {
     },
     inspector: {
       port: inspectorPort,
-      debugHost: debugHost,
       args: inspectorArgs
     }
   };
@@ -304,9 +297,7 @@ function openBrowserAndPrintInfo() {
   var url = inspector.buildInspectorUrl(
     'localhost',
     config.inspector.port,
-    config.subproc.debugPort,
-    null,
-    config.inspector.debugHost
+    config.subproc.debugPort
   );
 
   if (!config.options.cli) {
--- a/bin/node-debug.js
+++ b/bin/node-debug.js
@@ -27,6 +27,10 @@ var argvOptions = {
     type: 'number',
     description: 'Node/V8 debugger port (`node --debug={port}`)'
   },
+  'debug-host': {
+    type: 'string',
+    description: 'Host where the debugged app is running (`node-inspector --debug-host={url}`)'
+  },
   nodejs: {
     type: 'string',
     description: 'Pass NodeJS options to debugged process (`node --option={value}`)\n' +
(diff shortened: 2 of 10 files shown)


index.js
```diff
@@ -9,13 +9,16 @@ exports.buildInspectorUrl = buildInspectorUrl;
  * @param {number} inspectorPort as configured via --web-port
  * @param {number} debugPort as configured via --debug in the debugged app
  */
-function buildInspectorUrl(inspectorHost, inspectorPort, debugPort, fileToShow) {
+function buildInspectorUrl(inspectorHost, inspectorPort, debugHost, debugPort, fileToShow) {
```
bajtos: This is a public API, we need to preserve backwards compatibility. Please add a bit of argument-manipulation to support the old flavour `(inspectorHost, inspectorPort, debugPort, fileToShow)` too.

lib/DebuggerClient.js
```diff
@@ -21,8 +22,9 @@ function createFailingConnection(reason) {
  * @constructor
  * @param {number} debuggerPort
  */
-function DebuggerClient(debuggerPort) {
+function DebuggerClient(debuggerHost, debuggerPort) {
```
bajtos: Can we make this backwards-compatible by changing the order of parameters to `(debuggerPort, debuggerHost)`?

lib/PageAgent.js
```diff
@@ -46,6 +46,12 @@ extend(PageAgent.prototype, {
       this.emit('resource-tree');
     }.bind(this);
 
+    if (!this._debuggerClient.isHostMachine) {
+        return this._debuggerClient.once(
```
bajtos: Please add a comment explaining what's the purpose of this branch, i.e. skip building a list of source files from the filesystem because we are running on a different machine.

lib/config.js
```diff
@@ -81,6 +81,12 @@ var definitions = {
     convert: conversions.stringToInt,
     defaultValue: 5858
   },
+  'debug-host': {
+    desc: 'Host to run Node/V8 debugger.',
```
bajtos: `Host where the debugged app is running.`

lib/debug-server.js
```diff
@@ -62,8 +63,12 @@ DebugServer.prototype._getDebuggerPort = function(url) {
   return parseInt((/\?port=(\d+)/.exec(url) || [null, this._config.debugPort])[1], 10);
 };
 
-DebugServer.prototype._createSession = function(debugPort) {
-  return Session.create(debugPort, this._config);
+DebugServer.prototype._getDebuggerHost = function(url) {
+  return (/[\?\&]host=([0-9.]+)/.exec(url) || [null, this._config.debugHost])[1];
```
bajtos: This will support only IP addresses, hostnames like `localhost` won't work. Do we care?

lib/debugger.js
```diff
@@ -8,12 +8,13 @@ var Net = require('net'),
 /**
 * @param {Number} port
 */
-function Debugger(port){
+function Debugger(host, port){
```
bajtos: Ditto, can we change the argument order to `(port, host)` for backwards compatibility?

lib/debugger.js
```diff
@@ -146,6 +147,6 @@ Debugger.prototype.close = function() {
 * @param {Number} port
 * @type {Debugger}
 */
-module.exports.attachDebugger = function(port) {
-  return new Debugger(port);
+module.exports.attachDebugger = function(host, port) {
```
bajtos: Here too.

README.md
```diff
@@ -257,6 +257,7 @@ Command line options:
 ```
 --debug-brk, -b         Break on the first line (`node --debug-brk`) [default: true]
 --web-port, -p, --port  Node Inspector port (`node-inspector --web-port={port}`)
+--debug-port, -d        Node/V8 debugger ip address (`node-inspector --debug-host={port}`)
```
bajtos: `--debug-host`? (There is no one-letter shortcut registered AFAICT.)

README.md
```diff
@@ -305,6 +307,7 @@ List of predefined options:
 --help               |             | Print information about options
 --web-port           |    8080     | Port to host the inspector
 --web-host           |  127.0.0.1  | Host to listen on
+--debug-port         |  127.0.0.1  | Host where the debugging app runs on
```
bajtos: `--debug-host`

index.js
sparkleholic: Ok. I added a branch to get debugHost from the last argument without changing function scheme.
```
  if (buildInspectorUrl.arguments.length > 4 && buildInspectorUrl.arguments[4]) {
      parts.query['host'] = buildInspectorUrl.arguments[4];
  }
```

lib/PageAgent.js
sparkleholic: thanks. I added a comment as you mentioned.

lib/config.js
sparkleholic: Changed the description.

lib/debug-server.js
sparkleholic: I didn't catch it.
Can we use this instead ?
```
[\?\&]host=([0-9.a-zA-Z:/]+)/.exec(url)
```

lib/debugger.js
sparkleholic: changed `host` as a second argument.

README.md
sparkleholic: fixed a typo

sparkleholic:
I addressed the review. ready to review again.

bin/node-debug.js
```diff
@@ -27,6 +27,10 @@ var argvOptions = {
     type: 'number',
     description: 'Node/V8 debugger port (`node --debug={port}`)'
   },
+  'debug-host': {
+    type: 'string',
+    description: 'Host where the debugged app is running (`node-inspector --debug-host={url}`)'
+  },
```
bajtos: This doesn't make much sense to me, to be honest. `node-inspect` is intended to be run on the same machine where you 1) run the debugged app 2) open the browser. When debugging remotely, you have to start the debugged app manually. After that, there is little use in `node-inspect`, one can directly start `node-inspector`.

I am proposing to revert changes in this file (`bin/node-debug.js`).

If you disagree, then please check the output of `npm test`, there are several test failures caused by these changes.

https://travis-ci.org/node-inspector/node-inspector/jobs/183141504
```
 2) node-debug argument parser handles `app.js`:
      AssertionError: expected { Object (printScript, subproc, ...) } to deeply equal { Object (printScript, subproc, ...) }
      + expected - actual
       {
         "inspector": {
           "args": [
             "--web-port=8080"
      -      "--debug-host="
           ]
      -    "debugHost": ""
           "port": 8080
         }
         "printScript": true
         "subproc": {
      
      at Context.<anonymous> (test/node-debug.test.js:9:25)
  3) node-debug argument parser handles options:
      AssertionError: expected { Object (port, debugHost, ...) } to deeply equal { port: 10, args: [ '--web-port=10' ] }
      + expected - actual
       {
         "args": [
           "--web-port=10"
      -    "--debug-host="
         ]
      -  "debugHost": ""
         "port": 10
       }
      
      at Context.<anonymous> (test/node-debug.test.js:39:35)
  4) node-debug argument parser forwards unknown options to node-inspector:
      inspector args
      + expected - actual
         "--no-some-other-bool"
         "--some-string"
         "val"
         "--web-port=8080"
      -  "--debug-host="
       ]
      
      at Context.<anonymous> (test/node-debug.test.js:82:13)
```

index.js
```diff
@@ -15,8 +15,15 @@ function buildInspectorUrl(inspectorHost, inspectorPort, debugPort, fileToShow)
...
     port: inspectorPort,
     pathname: '/debug',
-    search: '?port=' + debugPort
+    query: {
+        port: debugPort
+    }
   };
+  // For backward compatibility, we'd like to keep arguments.
+  // But if arguments.length > 4, we take fifth argument as a debugHost.
+  if (buildInspectorUrl.arguments.length > 4 && buildInspectorUrl.arguments[4]) {
```
bajtos: To be honest, I haven't encountered this pattern before. We usually use `arguments` instead of `buildInspectorUrl.arguments`.

Is there any particular reason for using `arguments` instead of adding a new argument to the function signature?

```js
function buildInspectorUrl(inspectorHost, inspectorPort, debugPort, fileToShow, debugHost) {
  // ...
  if (debugHost) {
    parts.query['host'] = buildInspectorUrl.arguments[4];
  }
}
```

lib/PageAgent.js
```diff
@@ -46,6 +46,14 @@ extend(PageAgent.prototype, {
       this.emit('resource-tree');
     }.bind(this);
 
+    //skip building a list of source files from the filesystem
```
bajtos: `// skip...` (notice the missing space after double-slash, which is inconsistent with the next line)

lib/debug-server.js
```diff
@@ -62,8 +63,12 @@ DebugServer.prototype._getDebuggerPort = function(url) {
   return parseInt((/\?port=(\d+)/.exec(url) || [null, this._config.debugPort])[1], 10);
 };
 
-DebugServer.prototype._createSession = function(debugPort) {
-  return Session.create(debugPort, this._config);
+DebugServer.prototype._getDebuggerHost = function(url) {
+  return (/[\?\&]host=([0-9.a-zA-Z:/]+)/.exec(url) || [null, this._config.debugHost])[1];
```
bajtos: This will not support `?host=my-app.example.com` or even `?host=foo%20bar.com` (although I am not sure if a hostname is allowed to contain a character that needs escaping).


Here is the best solution to use:

```js
// at the top:
var url = require('url');
var querystring = require('querystring');

// here:
DebugServer.prototype._getDebuggerHost = function(urlString) {
  var parsedUrl = url.parse(urlString);
  var query = querystring.parse(parsedUrl.query);
  return query.host;
}
```

bin/node-debug.js
sparkleholic: I agree with your comment. I'll revert changes in `bin/node-debug.js`

index.js
sparkleholic: You're right. `arguments` seems proper here instead of ` buildInspectorUrl.arguments`. 
And I thought `debugPort` and `debugHost` should be adjacent so that the function arguments look intuitive, therefore I didn't add it to the function signature.
But now I changed my mind because it doesn't look different much that `arguments[4]` and `debugHost` as the last argument. I'll add `debugHost` to the function signature.

lib/PageAgent.js
sparkleholic: align format. Thanks.

lib/debug-server.js
sparkleholic: The code you have is a lot better ! :)

lib/session.js
```diff
@@ -10,7 +10,7 @@ var events = require('events'),
 ///////////////////////////////////////////////////////////
 // exports
 
-exports.create = function(debuggerPort, config) {
+exports.create = function(debuggerPort, debuggerHost, config) {
```
sparkleholic: I think I need to change this `debuggerHost` here to move into the last for the backward compatibility, right ? 
```
 exports.create = function(debuggerPort, config, debuggerHost) {
```

bajtos:
Landed and released as `node-inspector@0.7.5`. Enjoy :)

sparkleholic:
@bajtos Thanks and wish you happiness in the year to come. :)

PR #991, opened 2017-03-23
fix: upgrade v8-debug and v8-profiler, fix some test suites

hustxiaoc:


package.json
```diff
@@ -32,8 +32,8 @@
     "semver": "^4.3.4",
     "serve-favicon": "^2.1.1",
     "strong-data-uri": "^1.0.0",
-    "v8-debug": "~0.7.1",
-    "v8-profiler": "~5.6.0",
+    "v8-debug": "~1.0.0",
+    "v8-profiler": "~5.7.0",
```
bajtos: FWIW, I think it's better to use `^` operator for dependencies with post-1.0 versions. That way we don't have to updated these lines often.

```
"v8-debug": "^1.0.0",
"v8-profiler": "^5.7.0",
```

hustxiaoc: I'll update it in the next version.

Issue #993, opened 2017-03-30
Debugger not break at beginning when use --debug-brk


PrinC:
Node version 7.8.0
node-inspector version 1.0.0

When I use --debug-brk, with web debugger open,  js continue executing .
If I don't open node-inspector,  js executing will pause when using --debug-brk option

PrinC:
When I open node inspector web interface, console will report error with "Error: Debugger: Failed to resolve value by handle, ' #131# not found"

Muzietto:
Same as you, with node 6.9.4. It just started a couple of days ago. 
In my case the failed handle value is #3#

Muzietto:
If I completely kill the node.js cmd window and start a new one, then node-inspector 99% of the times works perfectly __for une run__.
This means that I have to kill the cmd window after each debugging run.

Muzietto:
Similar thing with intellij IDEA. __The first time__ I start a npm script containing a `node-debug --debug-brk` instruction, node-inspector works fine.
After that node-inspector again gives me the  "Error: Debugger: Failed to resolve value by handle, ' #3# not found" error.
But if I restart IDEA, the first run is gonna be OK.

Muzietto:
When an I get an error "Error: Debugger: Failed to resolve value by handle, ' #3# not found", I obviously kill the Chrome window with the red upper-right button.

- Most of the times, the node.js cmd window does not take notice, and I have to kill the running node process with a double CTRL-C; in those cases I know that node-inspector will fail at the next attempt.
- A few times, when I kill the Chrome window, the cmd window writes: "Error: This socket has been ended by the other party" and terminates my node app. __In this case, the next run with node-inspector will show no error__.

Muzietto:
__This problem is slowing down my development and has the highest priority in my agenda.
If any expert committer wants to take care of it, I am ready to give my full attention in providing testcases and trying solutions.__

hustxiaoc:
Sorry for the trouble, I'll fix it ASAP!

Muzietto:
__Great!!!__ If you need any more detail, just drop a note here.
Thank you.

hustxiaoc:
@Muzietto that's very nice of you, could you please check if this pr[https://github.com/node-inspector/node-inspector/pull/994] works ok with you?

Muzietto:
__Major improvement__, but still not 100% solved in my case:

- error "Error: Debugger: Failed to resolve value by handle, ' #3# not found" has disappeared
[...]

PR #994, opened 2017-04-09
fix: debug-brk not working and  failed to resolve value by handle

hustxiaoc:
fix https://github.com/node-inspector/node-inspector/issues/993

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ node_modules
 plugins/*
 v8.log
 .DS_Store
+.idea/
 /test/fixtures/work
 npm-debug.log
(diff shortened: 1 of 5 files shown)


Muzietto:
__Major improvement__, but still not 100% solved in my case:

- error "Error: Debugger: Failed to resolve value by handle, ' #3# not found" has disappeared
- instruction `--debug-brk` is still ignored, and I have to put a `debugger` instruction manually in my code
- after stopping at the `debugger` instruction, the debugger works flawlessly

I can live with that :-)

hustxiaoc:
`node-inspector@1.0.1` @Muzietto

PR #1002, opened 2017-04-12
doc: add doc of how to debug remote machine

hustxiaoc:
>#### Can I debug remotely?
>
>Yes. Node Inspector must be running on the same machine, but your browser can be anywhere.
>Just make sure port 8080 is accessible.
>
>And if Node Inspector is not running on your remote machine, you can also debug it as long as >your local machine can connect it.
>In this way, you must launch Node Inspector with `--no-inject` which means some features are not >supported such as profiling and consoling output inspection.

>So how to debug remote machine with your local Node Inspector?

>##### option 1
>$ node-inspector --debug-host 192.168.0.2 --no-inject
>then open the url
>http://127.0.0.1:8080/debug?port=5858
>
>##### option 2
>$ node-inspector  --no-inject
>then specify the remote machine address as a host parameter in the url
>e.g.) http://127.0.0.1:8080/debug?host=192.168.123.12&port=5858

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Node Inspector supports almost all of the debugging features of DevTools, includ
 
 ### Cool stuff
 * Node Inspector uses WebSockets, so no polling for breaks.
-* Remote debugging
+* Remote debugging and debugging remote machine.
 * [Live edit of running code](http://github.com/node-inspector/node-inspector/wiki/LiveEdit),
   optionally persisting changes back to the file-system.
 * Set breakpoints in files that are not loaded into V8 yet - useful for
@@ -125,6 +125,21 @@ When in doubt, refresh the page in browser
 
 Yes. Node Inspector must be running on the same machine, but your browser can be anywhere.
 Just make sure port 8080 is accessible.
+
+And if Node Inspector is not running on your remote machine, you can also debug it as long as your local machine can connect it.
+In this way, you must launch Node Inspector with `--no-inject` which means some features are not supported such as profiling and consoling output inspection.
+
+So how to debug remote machine with your local Node Inspector?
+
+##### option 1
+$ node-inspector --debug-host 192.168.0.2 --no-inject
+then open the url
+http://127.0.0.1:8080/debug?port=5858
+
+##### option 2
+$ node-inspector  --no-inject
+then specify the remote machine address as a host parameter in the url
+e.g.) http://127.0.0.1:8080/debug?host=192.168.123.12&port=5858
 
 #### How do I specify files to hide?
 
--- a/lib/DebuggerClient.js
+++ b/lib/DebuggerClient.js
@@ -1,7 +1,6 @@
 var extend = require('util')._extend;
 var EventEmitter = require('events').EventEmitter,
   inherits = require('util').inherits,
-  os = require('os'),
   DebugConnection = require('./debugger.js');
 
 function createFailingConnection(reason) {
@@ -50,20 +49,6 @@ Object.defineProperties(DebuggerClient.prototype, {
     get: function() {
       return this._conn.connected && !!this.target;
     }
-  },
-
-  isHostMachine: {
-      get: function() {
-          var addresses = ['localhost', '0.0.0.0'];
-          var ifaces = os.networkInterfaces();
-          for (var name in ifaces) {
-              var ifaceAddresses = ifaces[name].map( function(iface) {
-                  return iface['address'];
-              });
-              addresses = addresses.concat(ifaceAddresses);
-          }
-          return (addresses.indexOf(this._host) >= 0);
-      }
   }
 });
 
--- a/lib/InjectorClient.js
+++ b/lib/InjectorClient.js
@@ -4,9 +4,6 @@ var EventEmitter = require('events').EventEmitter,
     util = require('util'),
     debug = require('debug')('node-inspector:injector');
 
-// NM is NativeModule
-var FN_WITH_SCOPED_NM = 'Object.getOwnPropertyDescriptor(global, "console").get';
-
 /**
  * @param {{inject}} config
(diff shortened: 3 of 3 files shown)


hustxiaoc:
cc @sparkleholic

xmnboy:
I am having trouble figuring out how to make this work. Part of my confusion is understanding what is running where (remote versus local). 

> Define `remote` to mean _a system that is running the Node.js app which is to be debugged_ (for example, a headless IoT device that has Node.js installed and will be running the Node.js JavaScript app that we are trying to debug).

> Define `local` to mean _my laptop (i.e., workstation) on which I wish to run and view the debugger inside of my Chrome browser_ (so my local workstation, contains a keyboard, mouse and video screen and browser, but it is **NOT** running the Node.js JavaScript app that we wish to debug). 

Here's what I believe I should be doing:

1. On the `remote` machine, start my app with `node-debug`, for example:
   `$ node-debug --debug-brk main.js`

2. On the `local` machine, start `node-inspector`, for example:
    `$ node-inspector --debug-host 192.168.0.2 --no-inject`
(where `192.168.0.2` is the IP address of the `remote` machine)

3. On the `local` machine, copy `http://127.0.0.1:8080/?port=5858` into the Chrome browser (which is running on the `local` machine).

When I do that above, get some messages that imply a debug connection is almost working, but not quite. I see this sort of a message in the debug console that opens in my Chrome browser:

`Error: Cannot find module '/Users/pfische1/.nvm/versions/node/v6.9.1/lib/node_modules/node-inspector/lib/InjectorServer.js'`

and:

`Page.getResourceContent failed.
Error: ENOENT: no such file or directory, open '/home/ubuntu/node-debug-test/main.js'`

In fact, both files exist (on the `local` machine and on the `remote` machine, respectively). Perhaps I've got things backwards?

_Which commands need to be typed on which machine (`local` and `remote`) to make this work?_

XDKFeeds:
After some more probing and experiments I figured out what works for the situation I am describing above, which may be a different scenario than what you are trying to describe. 

The process that works for me is much simpler than what you describe in the docs:

1. At the console of the `remote` machine (e.g. an `ssh` session into `remote`), type:

> `$ node-debug --debug-brk --cli --web-host 192.168.0.2 my-app.js`

 Where `--debug-brk` is technically not required (I like to be safe), `--cli` makes sure the `node-debug` session does not attempt to automatically open a browser on the `remote` device, the IP address specified by `--web-host` is the `remote` device's IP address (the one to which you have a remote `ssh` console) and `my-app.js` is the name of the app you are going to start running and debug using `node-debug`. 

If the above is successful, you should see something like the following printed on the `remote` console, that indicates the `node-debug` session has begun:

> Node Inspector v1.1.1
> Visit http://192.168.0.2:8080/?port=5858 to start debugging.
> Debugging `my-app.js`
> 
> Debugger listening on [::]:5858

2. Open a browser (preferably, Chrome) on your `local` workstation and copy the URL provided above into your browser's address bar. 

If successful, you'll see an instance of remote Chrome DevTools running in the `local` browser and open to the first line of executable code in `my-app.js`. From here you can debug as normal. 

NOTE: if you are using any `console.log()` messages they should appear in the CDT console, but they will also appear in your `remote` device's `ssh` console. This is especially of note if your app ends with some `console.log()` messages, those messages will appear in the `ssh` console but may get dropped from the remote CDT console, due to a sudden loss of the debug connection.

xmnboy:
After some more probing and experiments I figured out what works for the situation I am describing above, which may be a different scenario than what you are trying to describe. 

The process that works for me is much simpler than what you describe in the docs:

1. At the console of the `remote` machine (e.g. an `ssh` session into `remote`), type:

> `$ node-debug --debug-brk --cli --web-host 192.168.0.2 my-app.js`

 Where `--debug-brk` is technically not required (I like to be safe), `--cli` makes sure the `node-debug` session does not attempt to automatically open a browser on the `remote` device, the IP address specified by `--web-host` is the `remote` device's IP address (the one to which you have a remote `ssh` console) and `my-app.js` is the name of the app you are going to start running and debug using `node-debug`. 

If the above is successful, you should see something like the following printed on the `remote` console, that indicates the `node-debug` session has begun:

> Node Inspector v1.1.1
> Visit http://192.168.0.2:8080/?port=5858 to start debugging.
> Debugging `my-app.js`
> 
> Debugger listening on [::]:5858

2. Open a browser (preferably, Chrome) on your `local` workstation and copy the URL provided above into your browser's address bar. 

If successful, you'll see an instance of remote Chrome DevTools running in the `local` browser and open to the first line of executable code in `my-app.js`. From here you can debug as normal. 

NOTE: if you are using any `console.log()` messages they should appear in the CDT console, but they will also appear in your `remote` device's `ssh` console. This is especially of note if your app ends with some `console.log()` messages, those messages will appear in the `ssh` console but may get dropped from the remote CDT console, due to a sudden loss of the debug connection.

hustxiaoc:
Ok, here we got two situations.
- 1. Start  app by running `node-debug --debug-brk main.js` on your remote machine, and then open your local browser with http://x.x.x.x.:8080/?port=5858.
- 2. Just make sure your app on the remote machine is lanuched with `--debug` , this doesn't require your app installing node-inspector, and then lanuch node-inspector with ` --debug-host 192.168.0.2 --no-inject` in your local machine.

So, basically you just need to start your app with `node --debug`, not `node-debug --debug-brk main.js` and  Node-inspector has nothing to do with your remote app in this case, just lanuch it on your local machine.

 Hope this will help, Thanks.

xmnboy:
Thanks for the clarification. I still cannot get anything to work using your second (`ii`) method, but that's fine because the first (`i`) method does exactly what I need. Including `node-inspector` and `node-debug` on my remote machine (the machine being debugged) is not an issue for me, so there's not much value for me in the second method.

